PR #32, opened 2023-05-19
Fix typos, spelling and make installation section more accurate

alamin655:
## What does this PR do?
Update README.md

so that the readme looks better.


neon-mmd (approved):
I think this looks much better :+1: , Thanks for putting the effort to make the Readme better for future visitors to the project.

neon-mmd:
I changed the PR title because I think the new title looks more appropriate, short, and descriptive of the changes you are making :smile: .

PR #40, opened 2023-05-22
Support the option to choose whether to add a random delay

XFFXFF:
## What does this PR do?
This pull request adds a new feature that allows users to choose whether or not to use websurfx in a production environment. With this update, you can set the `production_use` flag to `true` to introduce a random delay before sending requests to the search engine. Conversely, if you prefer to disable this functionality, you can simply change the option to `false`, like so:
```lua
production_use = false
```

This new feature provides users with greater flexibility and control over how they use the tool.

### DIFF
--- a/websurfx/config.lua
+++ b/websurfx/config.lua
@@ -19,3 +19,8 @@ theme = "simple" -- the theme name which should be used for the website
 
 -- Caching
 redis_connection_url = "redis://127.0.0.1:8082" -- redis connection url address on which the client should connect on.
+
+-- Aggregator
+aggregator = {
+  random_delay = false, -- whether to add random delay before sending the request to the search engine
+}
--- a/websurfx/config.lua
+++ b/websurfx/config.lua
@@ -20,6 +20,6 @@ theme = "simple" -- the theme name which should be used for the website
 -- Caching
 redis_connection_url = "redis://127.0.0.1:8082" -- redis connection url address on which the client should connect on.
 
-production_use = false -- whether to use production mode or not
+production_use = false -- whether to use production mode or not (in other words this option should be used if it is to be used to host it on the server to provide a service to a large number of users)
 -- if production_use is set to true
-  -- there will be a random delay before sending the request to the search engines, this is to prevent the search engines from blocking the ip address
+  -- There will be a random delay before sending the request to the search engines, this is to prevent DDoSing the upstream search engines from a large number of simultaneous requests.
--- a/websurfx/config.lua
+++ b/websurfx/config.lua
@@ -20,7 +20,6 @@ theme = "simple" -- the theme name which should be used for the website
 -- Caching
 redis_connection_url = "redis://127.0.0.1:8082" -- redis connection url address on which the client should connect on.
 
--- Aggregator
-aggregator = {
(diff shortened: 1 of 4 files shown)


websurfx/config.lua
```diff
@@ -19,3 +19,8 @@ theme = "simple" -- the theme name which should be used for the website
 
 -- Caching
 redis_connection_url = "redis://127.0.0.1:8082" -- redis connection url address on which the client should connect on.
+
+-- Aggregator
+aggregator = {
+  random_delay = false, -- whether to add random delay before sending the request to the search engine
```
XFFXFF, line 25: As of now, we are only deploying the tool locally, so by default, I have set the `random_delay` parameter to false. However, once we decide to deploy it publicly, we can create separate configurations for production and local environments

neon-mmd, line 25:
```suggestion
production_use = false
```

I feel that having an `aggregator` option with suboptions like `random_delay` will make it more ambiguous people will have questions like what is this `aggregator` option?? and Why is there a random delay ??. I think having a  `production_use` option will make it more self-documenting and less ambiguous for general users seeing the option itself they will get the idea like yes this option should only be set to true if it is to be used for large-scale use, also maybe in future we can think of adding more options under it but right now having `production_use` seems ok for now :slightly_smiling_face: .

neon-mmd (changes requested):

websurfx/config.lua
XFFXFF: That make sense!

XFFXFF:
> Also, one more thing, when you open up a PR fill the mandatory fields and all other fields that can be filled while others that are not necessary should be removed. This will allow us to better distinguish between spam PRs and legitimate PRs because most of the spam PRs never fill all the fields, I hope you got my point 😄 .

I see :)

websurfx/config.lua
```diff
@@ -20,7 +20,6 @@ theme = "simple" -- the theme name which should be used for the website
...
 redis_connection_url = "redis://127.0.0.1:8082" -- redis connection url address on which the client should connect on.
 
--- Aggregator
-aggregator = {
-  random_delay = false, -- whether to add random delay before sending the request to the search engine
-}
\ No newline at end of file
+production_use = false -- whether to use production mode or not
+-- if production_use is set to true
+  -- there will be a random delay before sending the request to the search engines, this is to prevent the search engines from blocking the ip address
```
neon-mmd, line 25:
```suggestion
production_use = false -- whether to use production mode or not (in other words this option should be used if it is to be used to host it on the server to provide a service to a large number of users)
-- if production_use is set to true
  -- There will be a random delay before sending the request to the search engines, this is to prevent DDoSing the upstream search engines from a large number of simultaneous requests.
```

I think mentioning `ip-blocking` will off-put many people from using the project and scare them instead of mentioning `DDoS` I feel would be more reasonable, also I think production should have some explanation as this makes it clear as to why that option has been added :smile: .

neon-mmd (changes requested):

neon-mmd (dismissed):
Ok looks great!! :smile: , let's merge it.

neon-mmd (approved):
Ok looks great!! :smile: , let's merge it.

PR #52, opened 2023-05-24
Ability to put config file on different paths

### LINKED ISSUE #44: Ability to put the config file under different paths
Add the ability to put the config files under different paths like `~/.config/websurfx/config.lua`, `/etc/xdg/websurfx/config.lua` and under the project folder path under `websurfx/` and also add code to properly handle each case like if the file is not available under `~/.config/websurfx/` then it fallback to using `/etc/xdg/websurfx/config.lua` if it is not available there then it fallback to using the project folder's config.lua which is located at `webusrfx/config.lua` and finally if it is not available there then it displays an appropriate error.

The code for this should look roughly like this:

``` rust 
fn handle_file_paths() -> Result<&'static str, Box<dyn Error>> {
    if Path::new(
        format!(
            "{}/.config/websurfx/config.lua",
            std::env::var("HOME").unwrap()
        )
        .as_str(),
    )
    .exists()
    {
        Ok("~/.config/websurfx/config.lua")
    } else if Path::new("/etc/xdg/websurfx/config.lua").exists() {
        Ok("/etc/xdg/websurfx/config.lua")
    } else if Path::new("./websurfx/config.lua").exists() {
        Ok("./websurfx/config.lua")
    } else {
        Err(format!("Config file not found!!").into())
    }
}
```

This will allow the project to be more easily deployable and installable on different `Linux distros` as well as it allows the project to be easily packaged for different `Linux distros`.

neon-mmd:
## What does this PR do?

This PR makes it possible to add the config file on paths other than the project folder (codebase).

This change allows the app to be packaged by Linux distros in their package repository for example this would allow it to be packaged for the [aur](https://aur.archlinux.org/) which makes it easier for users to install the app.

it can be tested by installing and running `websurfx` as mentioned in the docs and on the Readme and by checking whether on running it returns any error in the logs.

## Related issues

Closes #44


src/config_parser/parser.rs
```diff
@@ -52,4 +52,38 @@ impl Config {
...
+    /// 1. `~/.config/websurfx/` if it not present here then it fallbacks to the next one (2)
+    /// 2. `/etc/xdg/websurfx/config.lua` if it is not present here then it fallbacks to the next
+    ///    one (3).
+    /// 3. `websurfx/` (under project folder ( or codebase in other words)) if it is not present
+    ///    here then it returns an error as mentioned above.
+    fn handle_different_config_file_path() -> Result<String, Box<dyn std::error::Error>> {
+        if Path::new(
+            format!(
+                "{}/.config/websurfx/config.lua",
+                std::env::var("HOME").unwrap()
```
XFFXFF, line 71: Do we need to support deployment on Windows?  Windows might not have
 the `HOME` environment variable?

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "websurfx"
-version = "0.6.0"
+version = "0.8.0"
```
XFFXFF, line 3: Why is version 0.7.0 skipped?

src/config_parser/parser.rs
```diff
@@ -52,4 +52,38 @@ impl Config {
...
+        .exists()
+        {
+            Ok(format!(
+                "{}/.config/websurfx/config.lua",
+                std::env::var("HOME").unwrap()
+            ))
+        } else if Path::new("/etc/xdg/websurfx/config.lua").exists() {
+            Ok("/etc/xdg/websurfx/config.lua".to_string())
+        } else if Path::new("./websurfx/config.lua").exists() {
+            Ok("./websurfx/config.lua".to_string())
```
XFFXFF, line 84: I noticed that `websurfx/config.lua` is repeated multiple times in the code. This could potentially lead to errors, particularly if we make a typo like `websurrfx/config.lua` somewehre. To avoid such issues, I suggest using a variable to represent `websurfx/config.lua` and utilizing it instead of a raw string.

XFFXFF (commented):
nice!

some comments

src/config_parser/parser.rs
neon-mmd: I think for right now we will not provide support for Windows because most of the users of this project will be Linux users because most of the users will be privacy-minded individuals so they will tend to use Linux for the project so for now `No` but in future, if we do get a lot of requests from users then we will consider adding support for `windows` and `macOS`. :slightly_smiling_face:

Cargo.toml
neon-mmd: Actually, I forgot to ask you last time in the previous PR to bump the version number, My bad. so because of that, we have a release with v0.7.0 but with a `Cargo.toml` of 0.6.0 which is very bad, also being a maintainer if I don't take care of such small mistakes, it is very shameful that's why, from now on I am taking care of this as well :slightly_smiling_face:  .

src/config_parser/parser.rs
neon-mmd: Yes, you are right :+1: :smile: . I will make the change right away, just give me a few seconds :slightly_smiling_face: .

XFFXFF (approved):

PR #53, opened 2023-05-24
Move docs from wiki to docs folder in the codebase

### LINKED ISSUE #47: Should we move the documentation from the wiki section to the "docs" folder in the codebase??
I think having the documentation under the wiki section makes it go unnoticed most of the time because of that sometimes it remains out of sync with the changes that are being made which is bad as you know so I feel having it under the `docs` folder will make it more noticeable. What do you say @XFFXFF @alamin655 ??.

neon-mmd:
## What does this PR do?

This PR moves the docs from the wiki section of the repository to the docs folder in the codebase and adds an image in the homepage of the docs.

This change makes handling docs easier and makes it more noticeable for contributions.

## Related issues

Closes #47


XFFXFF (approved):
nice!

PR #55, opened 2023-05-24
Ability to have themes on different paths

neon-mmd:
## What does this PR do?

This PR makes it possible to add the themes folder on paths other than the project folder (codebase).

This change allows the app to be packaged by Linux distros in their package repository for example this would allow it to be packaged for the [aur](https://aur.archlinux.org/) which makes it easier for users to install the app.

it can be tested by installing and running `websurfx` as mentioned in the docs and on the Readme and by checking whether on running it returns any error in the logs.

## Related issues

Closes #52

### DIFF
--- a/src/server/routes.rs
+++ b/src/server/routes.rs
@@ -7,8 +7,8 @@ use std::fs::read_to_string;
 use crate::{
     cache::cacher::RedisCache,
     config_parser::parser::Config,
+    handler::public_path_handler::handle_different_public_path,
     search_results_handler::{aggregation_models::SearchResults, aggregator::aggregate},
-    theme_handler::theme_path_handler::handle_different_theme_path,
 };
 use actix_web::{get, web, HttpRequest, HttpResponse};
 use handlebars::Handlebars;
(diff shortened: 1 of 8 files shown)


src/server/routes.rs
```diff
@@ -146,7 +147,8 @@ pub async fn search(
 /// Handles the route of robots.txt page of the `websurfx` meta search engine website.
 #[get("/robots.txt")]
 pub async fn robots_data(_req: HttpRequest) -> Result<HttpResponse, Box<dyn std::error::Error>> {
-    let page_content: String = read_to_string("./public/robots.txt")?;
+    let page_content: String =
+        read_to_string(format!("{}/robots.txt", handle_different_theme_path()?))?;
```
XFFXFF, line 151: I may seem confusing as to why the `robots.txt` file is located under the theme path. Since the theme path is essentially the public path, I think it could be better to consider a function name such as `handle_differenct_public_path` instead of `handle_different_theme_path`. What do you say? @neon-mmd

XFFXFF (commented):
Looks great! Only one comment

src/server/routes.rs
neon-mmd: My thinking was that the name `theme` would be more appropriate while writing the code but now I think you are right, I will make the change right away :slightly_smiling_face: .

XFFXFF (approved):

PR #60, opened 2023-05-26
fix issue: #56 removed the meta seach engine link

### LINKED ISSUE #56: Readme Revision
Revise the readme to keep it on par with the current changes. These include adding and changing the following things:
- [x] Remove unnecessary shields from the `Readme`
- [x] Update the first preview image in the readme if possible replace the images with a `Gif`
- [x] Fix `Back to Top` button by updating the link to where it points.
- [x] Update the `Why Websurfx` section content under the `FAQ` section from this:

> The primary purpose of the Websurfx project is to create a fast, secure, and privacy-focused [meta search engine](https://en.wikipedia.org/wiki/Metasearch_engine). While there are numerous meta search engines available, not all of them guarantee the security of their search engine, which is critical for maintaining privacy. Memory flaws, for example, can expose private or sensitive information, which is never a good thing. Rust is used to write Websurfx, which ensures memory safety and removes such issues. Many meta search engines also lack important features like advanced picture search, which is required by many graphic designers, content providers, and others. Websurfx attempts to improve the user experience by providing these and other features, such as proper NSFW blocking.

to:

> The primary purpose of the Websurfx project is to create a fast, secure, and privacy-focused meta-search engine. While there are numerous meta-search engines available, not all of them guarantee the security of their search engine, which is critical for maintaining privacy. Memory flaws, for example, can expose private or sensitive information, which is never a good thing. Also, there is the added problem of Spam, ads, and unorganic results which most engines don't have the full-proof answer to it till now but with Websurfx I finally put a full stop to this problem, also, Rust is used to write Websurfx, which ensures memory safety and removes such issues. Many meta-search engines also lack important features like advanced picture search, which is required by many graphic designers, content providers, and others. Websurfx attempts to improve the user experience by providing these and other features, such as proper NSFW blocking and Micro-apps or Quick results (like providing a calculator, currency exchanges, etc in the search results).

sathiyaIbe:
## What does this PR do?

Removed the meta-search engine link

The issues were opened in the readme file is this needed to avoid an external link

## Related issues

#56


neon-mmd:
Ok, I love the changes you made and I would be very glad to merge them but can you please explain in brief (adding a little more context) why did you make the change ?? Like what was the reason behind for example `Did you want to reduce clutter?` or `Remove the link because it was unnecessary? Explaining the reason why it was so?` or `something else?`

sathiyaIbe:
yeah that meta search engine text contains some link and that goes to some external page so avoid this I had removed the background link as well as in that paragraph again the same word was mentioned without the background link so its seems to be unwanted so that's why I made these changes

neon-mmd (approved):
Ok nice!! :+1: , let's merge it :slightly_smiling_face: .

sathiyaIbe:
Thank you for your response really super fast

PR #61, opened 2023-05-26
UPDATED: Removed Deprecated API

LX5321:
## What does this PR do?

Updates the API for capturing the `Enter` keypress from keyboard.

Fixes code in accordance to CWE-477: Use of Obsolete Function.

[https://cwe.mitre.org/data/definitions/477](CWE-477)

### DIFF
--- a/public/static/index.js
+++ b/public/static/index.js
@@ -4,7 +4,7 @@ function search_web() {
 }
 
 search_box.addEventListener('keyup', (e) => {
-  if (e.keyCode === 13) {
+  if (e.key === 'Enter') {
     search_web()
   }
 })


neon-mmd:
Ok, the code looks great!! :+1: Thanks for the fix :slightly_smiling_face: I would love to have this merged but if you don't mind can you please reopen this PR by adding all the changes you made to the `rolling branch` of the forked project and then changing the base branch to `rolling` instead of `master`.

So that it instead of this:
image

It shows something like:
`neon-mmd:rolling` from `LX5321:rolling`

LX5321:
@neon-mmd Sure, not a problem 👍🏼

PR #69, opened 2023-05-27
Centered About page content

### LINKED ISSUE #59: Make about page centered
Put some margins on the right and left of the content of the `About Page` to make it centered on the website to make it visually look better by using percentage values for the margins instead of using fixed values like `rem`, `em`, `px`, etc. 

image

The target file that needs to be changed is located under `public/static/themes` and it is called `simple.css`.

sam-sunder:
This PR fixes https://github.com/neon-mmd/websurfx/issues/59 centering about page content

This is an UI fix

## Related issues
Closes #59

### DIFF
--- a/public/static/themes/simple.css
+++ b/public/static/themes/simple.css
@@ -296,5 +296,8 @@ footer {
 
   .about-container h3{
     font-size: 1.5rem;
+  }
+
+  .about-container {
     width: 80%;
   }
--- a/public/static/themes/simple.css
+++ b/public/static/themes/simple.css
@@ -296,4 +296,5 @@ footer {
 
   .about-container h3{
     font-size: 1.5rem;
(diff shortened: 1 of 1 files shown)


sam-sunder:
image

neon-mmd:
I don't why but the `About page` content doesn't seem to be centered, Maybe you made a slight error or mistake in the code for example `you wanted to add something else but mistakenly added something other than what was intended`, I would suggest reviewing the code once more :slightly_smiling_face: .

image

sam-sunder:
Thanks, I've reviewed code and found the mistake. It's ready to be reviewed by you now 😊

neon-mmd (approved):
Now it looks perfect :+1: !! I will merge this change right away :slightly_smiling_face: .

PR #70, opened 2023-05-27
Provide an option to enable or disable logs

### LINKED ISSUE #46: Provide logging option in the config
Provide a `logging` option that when set to true enables logging in standard output but when set to false disables it. This option is important for privacy reasons as this will give users the ability to select servers that have these options set to false which will give them a sense of relief.

``` lua
logging = true
```

neon-mmd:
## What does this PR do?

This PR introduces a new option to enable or disable logs.

This change allows users to disable logs to improve their privacy.

it can be tested by installing and running websurfx as mentioned in the docs and on the Readme and by checking whether on running it returns any errors.

## Author's checklist

- [x] Add a new option named `logging` in the config file.
- [x] Add the code to parse the new option.
- [x] Add a condition-based code to only enable logging if it is set to true.  

## Related issues

Closes #46

### DIFF
--- a/src/bin/websurfx.rs
+++ b/src/bin/websurfx.rs
@@ -5,7 +5,6 @@
 
 use std::net::TcpListener;
 
-use env_logger::Env;
 use websurfx::{config_parser::parser::Config, run};
 
 /// The function that launches the main server and registers all the routes of the website.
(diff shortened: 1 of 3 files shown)


xffxff (approved):
nice!

PR #79, opened 2023-05-29
Github CI to automate the generation of contributors list

### LINKED ISSUE #78: Automate Contributor list generation
At present the contributor list is being provided by contrib.rocks but this solution seems to have problems like it doesn't update that often which makes the contributor list under the `credits` section of `Readme` lag behind as shown in the image below. Provide a way to automate the GitHub contributors list using an appropriate Github action that runs every day on a fixed schedule for example at 10 pm every day.

Here you can see the actual number of contributors is 8 but in the Readme, it shows only 5 contributors:

image

image

`Note:` The Github you choose should provide the ability to provide rounded avatars in the contributors' list.

neon-mmd:
## What does this PR do?

This PR adds a GitHub CI to automate the generation of the contributors' list.

This change is essential as this way is a better solution to provide a contributors' list as previously seemed to lag behind in showing the actual numbers of contributors (as shown in the images below)  in the contributors under the `Credits` section of the `Readme`. 

image

image

## Related issues

Closes #78

### DIFF
--- a/.github/workflows/contributors.yml
+++ b/.github/workflows/contributors.yml
@@ -1,0 +1,16 @@
+name: Contributors
+on:
+  schedule:
+    - cron: '0 1 * * *' # At 01:00 on every day.
+  push:
+    branches:
+      - rolling
+jobs:
+  contributors:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: wow-actions/contributors-list@v1
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          round: true
+          svgPath: ../../images/contributors_list.svg


.github/workflows/contributors.yml
```diff
@@ -0,0 +1,16 @@
...
+      - rolling
+jobs:
+  contributors:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: wow-actions/contributors-list@v1
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          round: true
+          svgPath: ../../images/contributors_list.svg
```
xffxff, line 16: I can't find where we use this svg. Do we need to use it in the Readme.md?

neon-mmd: Yes :slightly_smiling_face: , we need to use the generated `svg file` in the `Readme.md` if you want to like to know more about the GitHub action you can check this [out](https://github.com/marketplace/actions/contributors-list) :slightly_smiling_face: .

xffxff: I see ~

xffxff (approved):
nice!

PR #80, opened 2023-05-29
Provide a debug option in the config

### LINKED ISSUE #45: Provide a debug option in the config
Provide a debug option in the config file. This option when set to true will give a verbose log and also it will disable random delays.

``` lua
debug = true
```

neon-mmd:
## What does this PR do?

This change provides a new debug option in the config file. 

This change allows anyone working on the project or testing the project to view in-depth the different happening in the app. 

it can be tested by installing and running websurfx as mentioned in the docs and on the Readme and by checking whether on running it returns any errors.

## Author's checklist

- [x] Add a new option named `debug` in the config file.
- [x] Add the code to parse the new option.
- [x] Add a condition-based code to only enable `debug mode` if it is set to true.
- [x] Add a condition-based code to only disable random delay when `debug` option is set to true.

## Related issues

Closes #45

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -125,7 +125,7 @@ dependencies = [
  "actix-utils",
  "futures-core",
  "futures-util",
- "mio 0.8.6",
+ "mio 0.8.7",
  "num_cpus",
  "socket2",
  "tokio 1.28.2",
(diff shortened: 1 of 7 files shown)


xffxff (approved):
Looks good

PR #84, opened 2023-05-30
Provide appropriate permissions for contributors' list generation ci

neon-mmd:
## What does this PR do?

This PR adds necessary permissions to the contributions' list autogeneration ci by updating the respective yaml file based on the issue [https://github.com/wow-actions/contributors-list/issues/4](https://github.com/wow-actions/contributors-list/issues/4).

This change fixes the contributors' list generation ci proposed in #78.


xffxff (approved):
Not sure if it works, but let's give it a try and see what happens

neon-mmd:
> Not sure if it works, but let's give it a try and see what happens

I think it worked, we now have that green checkmark showing everything is working fine :slightly_smiling_face: , so now we just need to wait for the action to run at the specified time as specified in the cron value and let's see if everything goes according to the plan :slightly_smiling_face: .

Issue #37, opened 2023-05-21
Save all the configuration setting from the setting UI to a cookie when save button is pressed


neon-mmd:
Add functionality to save all the configurations configured by the user as a cookie when the `save` button is pressed on the settings page.

neon-mmd:
This issue has been taken care of by PR https://github.com/neon-mmd/websurfx/pull/88, so, therefore, I mark this issue as completed :slightly_smiling_face: .

PR #88, opened 2023-05-31
Provide a Settings Page

### LINKED ISSUE #36: Create a settings page to allow users to select the search engines from the UI
Create an Ui to select different configuration options like themes, colorschemes and different engines to choose from with the help of checkboxes and also provide a `select all` and `deselect all` button to select/deselect all engines at once.

sam-sunder:
## What does this PR do?

This PR implements the settings page according to the requirements provides in issue #36 and #37  .

This PR is important as it implements a settings page for the websurfx to allow users to easily interact and choose between various options from the UI itself.

## Related issues

Closes #36 and #37

## Preview
image


alamin655 (changes requested):
Thank you, @sam-sunder, for taking the time to provide this PR on the basis of the requirements listed in issue #36 :slightly_smiling_face: .

I love the changes you made :+1: the UI looks cool :rocket: and I am eager to merge this PR but I have a few concerns and for clarity reasons, I would list them below:
- I would prefer to have all the settings put together as a cookie so that you can store it on the user's system and whenever he will use the app then all the settings from the cookie can be easily loaded from the user's machine to the app.
- Also I would like you to use JS less because it tends to have a lot of vulnerabilities and tends to be unsafe language as well as it is considered very bad for privacy. 
- Also I see that you have added some colorscheme variables which may cause issues, specifically the variables you used, `--1_1: #4f5169;` and `--3_1: #7ce073;`. To avoid such issues, use the recommended variables that we defined and use them consistently throughout the UI. This will ensure that the color values used in the UI are compatible with all other colorschemes.
- Also the formatting of the code seems a little inconsistent and would like you to look into this regard as well. 
- Also for the engine selection menu I feel it would be better to have something similar to the image shown below :point_down:.

image

I hope you will not mind making the following changes :slightly_smiling_face: .

sam-sunder:
Updated code as per your review

Here's Preview of UI: [Screenshot](https://user-images.githubusercontent.com/89068308/243071921-34c71f28-9520-44a8-9d3e-926955494d00.png)

Here are some updates:
- Added Toggle options to select search engines.
- Added a cookie which saves the settings.
- Refactored JS to increase readability.
- Added JS of UI components separately which can be used universally at any page.
- Removed loading css themes and colors scheme by default, from now theme, color scheme will be loaded from javascript as per the user settings.
- Removed additional variables from color schemes.
- This PR also closes #37

neon-mmd:
I would prefer to reduce the amount of JS in the code like make it so that it is used only when it is real necessity otherwise CSS is enough for the work :slightly_smiling_face: , is there any way to do so @sam-sunder ??

sam-sunder:
I'm not sure
But this js is only for UI components it won't effect any privacy and speed of the application, it also include saving of settings to cookie and loading settings from cookie. So we need JS. ✌️

neon-mmd:
> I'm not sure But this js is only for UI components it won't affect any privacy and speed of the application, it also includes saving settings to cookie and loading settings from the cookie. So we need JS. :v: 

Ok, I see :eyes:, then that's great!! :rocket:  :slightly_smiling_face: but just one more thing can you please make the comments above functions a little more descriptive and more explaining as to what the function does?? I would suggest using [JS convention](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Writing_style_guide/Code_style_guide/JavaScript) for this :slightly_smiling_face: .

neon-mmd (approved):
Ok, this looks great!! :+1:, let's merge this in :rocket: :slightly_smiling_face: .

PR #90, opened 2023-06-01
Put pre-commit checks to ensure better code quality

neon-mmd:
## What does this PR do?

This PR adds pre-commit checks using rusty-hook and adds a stylelint config file.

This change ensures that any code before making it through the PR is consistent and has better quality.

## Author's checklist

- [x] Update the `cargo.toml` file with the rusty-hook dependency by putting it under the dev dependency
- [x] Add `.rusty-hook.toml` file with various checks.
- [x] Add `.stylelintrc.json` to configure stylelint with appropriate linting options.
- [x] Update .gitignore file.

### DIFF
--- a/.rusty-hook.toml
+++ b/.rusty-hook.toml
@@ -1,0 +1,5 @@
+[hooks]
+pre-commit = "cargo test && cargo fmt -- --check && cargo clippy && stylelint ./public/static/themes/*.css ./public/static/colorschemes/*.css ./public/static/*.js" 
+
+[logging]
+verbose = true
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 /target
-
+package.json
+package-lock.json
 dump.rdb
+.vscode
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,0 +1,13 @@
+{  
+ "extends": "stylelint-config-standard",
+  "rules": {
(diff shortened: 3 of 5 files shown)


.rusty-hook.toml
alamin655, line 1: We should consider separating pre-commit checks:

- **Improved readability**: Each check is clearly separated, making it easier to understand and modify specific checks as needed.
- **Flexibility in customization**: This allows for easy customization or extension of the checks to meet different project requirements. For example, specific `stylelint` commands can be added or removed based on desired file patterns or directories.

```toml
[hooks]
pre-commit = [
  "cargo test",
  "cargo fmt -- --check",
  "cargo clippy",
  "stylelint ./public/static/themes/*.css",
  "stylelint ./public/static/colorschemes/*.css",
  "stylelint ./public/static/*.js"
]

[logging]
verbose = true
```

neon-mmd: Yes, I think this looks better :+1:, but there is a cache with it you can't add these as a list because `rusty-hook` doesn't support this and it gives an error if you do so :slightly_smiling_face: .

alamin655: I apologize for not checking it earlier. It was my mistake. In that case, we can proceed with the code you provided and merge it.

alamin655 (approved):
LGTM, let's get it merged.

PR #105, opened 2023-06-15
Bump the app version number

neon-mmd:
## What does this PR do?

This PR bumps the app version to v0.12.1.

This change is important to keep the app version in sync with the releases.


alamin655 (approved):
Thanks :slightly_smiling_face: , Let's get it merged.

PR #110, opened 2023-06-18
Improve the code associated with the settings page

neon-mmd:
## What does this PR do?

This PR reworks the previous PR #88 and reduces the amount of JS used and removes the unnecessary search link from the navbar while keeping the same look and feel as done and proposed in PR #88 by @sam-sunder .

This change is essential as it reduces the amount of JS used, which is necessary to ensure the privacy of the users.

It can be tested by installing and running Websurfx as mentioned in the docs and on the Readme and by launching the browser and going to the settings and by clicking and testing on every button if it works.

```
http://127.0.0.1:8080/settings
```

## Author's checklist

- [x] Reduce the amount of JS used.
- [x] Remove the unnecessary search link from the navbar.
- [x] Add documentation to code.
- [x] Bump the app version.  

## Related issues

#36 and #37

## Additional Notes

Special thanks to @sam-sunder for working on the settings page, contributing a wonderful settings page and doing all the heavy lifting work. Hoping to see you seen with wonderful PR like #88 in the future as well :slightly_smiling_face: .


public/static/cookies.js
alamin655, line 1: The code looks good as it is, but here are a few suggestions for improvement:

1. Instead of using `querySelector` to select the input element, we can use `getElementById` to select the element by its ID. This is faster and more efficient.

2. We can use a ternary operator to simplify the code that sets the input value. Here's an example:
```js
document.getElementById('cookies-input').value = cookie !== '' ? cookie : 'No cookies have been saved on your system';
```
3. We can add error handling to the code to handle cases where the cookie cannot be decoded. Here's an example:
```js
try {
  let cookie = decodeURIComponent(document.cookie);
  document.getElementById('cookies-input').value = cookie !== '' ? cookie : 'No cookies have been saved on your system';
} catch (error) {
  console.error('Error decoding cookie:', error);
  document.getElementById('cookies-input').value = 'Error decoding cookie';
}
```

public/templates/navbar.html
```diff
@@ -1,6 +1,6 @@
 <nav>
-  <ul>
-    <li><a href="about">about</a></li>
-    <li><a href="settings">settings</a></li>
-  </ul>
+    <ul>
```
alamin655, line 2:
```suggestion
    <ul>
        <li><a href="search">search</a></li>
```

public/static/settings.js
```diff
@@ -0,0 +1,83 @@
...
+// options in the settings page under the tab engines.
+function toggleAllSelection() {
+  document
+    .querySelectorAll('.engine')
+    .forEach(
+      (engine_checkbox) =>
+        (engine_checkbox.checked =
+          document.querySelector('.select_all').checked)
+    )
+}
```
alamin655, line 11: The function, `toggleAllSelection`, uses `querySelectorAll` which can be slow if there are many elements to search through. It also uses `forEach` which can be slow if there are many elements to iterate through. A more performant solution would be to use a `for loop` instead of `forEach`.

```suggestion
function toggleAllSelection() {
  const engineCheckboxes = document.querySelectorAll('.engine');
  const selectAllCheckbox = document.querySelector('.select_all');
  for (let i = 0; i < engineCheckboxes.length; i++) {
    engineCheckboxes[i].checked = selectAllCheckbox.checked;
  }
}
```

public/static/settings.js
```diff
@@ -0,0 +1,83 @@
...
+    .querySelectorAll('.tab')
+    .forEach((tab) => tab.classList.remove('active'))
+  document
+    .querySelectorAll('.btn')
+    .forEach((tab) => tab.classList.remove('active'))
+  current_tab.classList.add('active')
+  document
+    .querySelector(`.${current_tab.innerText.toLowerCase().replace(' ', '_')}`)
+    .classList.add('active')
+}
```
alamin655, line 26: The function, `setActiveTab`, uses `querySelectorAll` which can be slow if there are many elements to search through. It also uses `forEach` which can be slow if there are many elements to iterate through. A more performant solution would be to use a `for loop` instead of `forEach`.
```suggestion
function setActiveTab(currentTab) {
  const tabs = document.querySelectorAll('.tab');
  const buttons = document.querySelectorAll('.btn');
  for (let i = 0; i < tabs.length; i++) {
    tabs[i].classList.remove('active');
  }
  for (let i = 0; i < buttons.length; i++) {
    buttons[i].classList.remove('active');
  }
  currentTab.classList.add('active');
  const tabName = currentTab.innerText.toLowerCase().replace(' ', '_');
  document.querySelector(`.${tabName}`).classList.add('active');
}
```

public/static/settings.js
```diff
@@ -0,0 +1,83 @@
...
+    cookie_dictionary
+  )}; expires=${expiration_date.toUTCString()}`
+
+  document.querySelector('.message').innerText =
+    '✅ The settings have been saved sucessfully!!'
+
+  setTimeout(() => {
+    document.querySelector('.message').innerText = ''
+  }, 10000)
+}
```
alamin655, line 58: The function, `setClientSettings`, uses `querySelectorAll` which can be slow if there are many elements to search through. It also uses `forEach` which can be slow if there are many elements to iterate through. A more performant solution would be to use a `for loop` instead of `forEach`.
```suggestion
function setClientSettings() {
  const cookieDictionary = {};
  const selectTags = document.querySelectorAll('select');
  for (let i = 0; i < selectTags.length; i++) {
    const selectTag = selectTags[i];
    if (selectTag.name === 'themes') {
      cookieDictionary['theme'] = selectTag.value;
    } else if (selectTag.name === 'colorschemes') {
      cookieDictionary['colorscheme'] = selectTag.value;
    }
  }
  const engines = [];
  const engineCheckboxes = document.querySelectorAll('.engine');
  for (let i = 0; i < engineCheckboxes.length; i++) {
    const engineCheckbox = engineCheckboxes[i];
    if (engineCheckbox.checked === true) {
      engines.push(engineCheckbox.parentNode.parentNode.innerText.trim());
    }
  }
  cookieDictionary['engines'] = engines;
  const expirationDate = new Date();
  expirationDate.setFullYear(expirationDate.getFullYear() + 1);
  document.cookie = `appCookie=${JSON.stringify(cookieDictionary)}; expires=${expirationDate.toUTCString()}`;
  document.querySelector('.message').innerText = '✅ The settings have been saved sucessfully!!';
  setTimeout(() => {
    document.querySelector('.message').innerText = '';
  }, 10000);
}
```

public/static/settings.js
```diff
@@ -0,0 +1,83 @@
...
+      (item) => {
+        if (item.href.includes('static/themes')) {
+          item.href = `static/themes/${cookie_value['theme']}.css`
+        } else if (item.href.includes('static/colorschemes')) {
+          item.href = `static/colorschemes/${cookie_value['colorscheme']}.css`
+        }
+      }
+    )
+  }
+}
```
alamin655, line 83: The function, `getClientSettings`, uses `decodeURIComponent` twice on the same cookie string. It also uses `querySelectorAll` which can be slow if there are many elements to search through. It also uses `forEach` which can be slow if there are many elements to iterate through. A more performant solution would be to use a `for loop` instead of `forEach` and to only call `decodeURIComponent` once.
```suggestion
function getClientSettings() {
  const cookie = decodeURIComponent(document.cookie);
  if (cookie !== '') {
    const cookieValue = cookie.split(';').reduce((acc, item) => {
      const [key, value] = item.split('=');
      acc[key.trim()] = JSON.parse(decodeURIComponent(value));
      return acc;
    }, {});
    const links = document.querySelectorAll('link');
    for (let i = 0; i < links.length; i++) {
      const link = links[i];
      if (link.href.includes('static/themes')) {
        link.href = `static/themes/${cookieValue['theme']}.css`;
      } else if (link.href.includes('static/colorschemes')) {
        link.href = `static/colorschemes/${cookieValue['colorscheme']}.css`;
      }
    }
  }
}
```

public/static/cookies.js
```diff
@@ -0,0 +1,17 @@
+// This function is executed when any page on the website finsihes loading and
```
alamin655, line 1:
```suggestion
// This function is executed when any page on the website finishes loading and
```

public/static/cookies.js
```diff
@@ -0,0 +1,17 @@
...
+    let cookie = decodeURIComponent(document.cookie)
+    if (cookie !== '') {
+      document.querySelector('.cookies input').value = cookie
+    } else {
+      document.querySelector('.cookies input').value =
+        'No cookies has been saved on your system'
+    }
+  },
+  false
+)
```
alamin655:
```suggestion
document.addEventListener(
  'DOMContentLoaded',
  () => {
    const cookie = decodeURIComponent(document.cookie)
    if (cookie === '') {
      document.querySelector('#cookies-input').value =
        'No cookies have been saved on your system'
    } else {
      document.querySelector('#cookies-input').value = encodeURIComponent(cookie)
    }
  },
  false
)
```

public/templates/cookies_tab.html
```diff
@@ -0,0 +1,12 @@
+<div class="cookies tab">
+    <h1>Cookies</h1>
+    <p class="description">
+        This is the cookies are saved on your system and it contains the preferences
```
alamin655, line 4:
```suggestion
        This is where cookies are saved on your system. They contain the preferences you chose in the settings page.
```

public/templates/cookies_tab.html
```diff
@@ -0,0 +1,12 @@
...
+        you chose in the settings page
```
alamin655, line 5:
```suggestion
        This is where cookies are saved on your system. They contain the preferences you chose in the settings page.
```

public/templates/cookies_tab.html
```diff
@@ -0,0 +1,12 @@
...
+    </p>
+    <input type="text" name="cookie_field" value="" readonly />
```
alamin655, line 7:
```suggestion
    <input type="text" name="cookie_field" value="" readonly>
```

public/templates/cookies_tab.html
```diff
@@ -0,0 +1,12 @@
...
+    <p class="description">
+        The cookies stored are not used by us for any malicious intend or for
+        tracking you in any way.
```
alamin655, line 10:
```suggestion
        The cookies stored are not used by us for any malicious intent or for tracking you in any way.
```

public/templates/engines_tab.html
```diff
@@ -0,0 +1,31 @@
+<div class="engines tab">
+    <h3>select search engines</h3>
+    <p class="description">
+        Select the search engines from the list of engines that you want results
+        from
```
alamin655, line 5:
```suggestion
        Select the search engines from the list of engines that you want results from
```

public/templates/engines_tab.html
```diff
@@ -0,0 +1,31 @@
...
+            </label>
+            Select All
+        </div>
+        <hr />
+        <div class="toggle_btn">
+            <label class="switch">
+                <input type="checkbox" class="engine" />
+                <span class="slider round"></span>
+            </label>
+            Duckduckgo
```
alamin655, line 21:
```suggestion
            DuckDuckGo
```

alamin655 (changes requested):
I made some recommendations and changes.

public/static/cookies.js
neon-mmd: Thanks for the suggestion, This looks awesome!! :rocket: I would make this change right away :slightly_smiling_face: .

public/templates/navbar.html
neon-mmd: I like your suggestions and I really appreciate it, thanks once again for suggesting this. The problem with this change is that the search page takes some query parameters and navigating to that page without any query parameters would result in an error which is not very useful. My suggestion is to have a logo in the top left corner of the navbar which, when clicked, takes the user to the page of the website, What do you say?

alamin655: Yeah, we can add a logo, but for now, until the logo is added, we can use this, and I didn't have any issue using it.

public/static/cookies.js
neon-mmd: Thanks for the suggestion, but I think we can add it as a new issue with a good-first-issue label, this allows new contributors to get involved in the project and enable them to add value to the project. What do you say??

public/templates/engines_tab.html
neon-mmd: Thanks for the suggestion, but I think we can add it as a new issue with a good-first-issue label, this allows new contributors to get involved in the project and enable them to add value to the project. What do you say??

public/static/cookies.js
alamin655: Yeah good idea.

neon-mmd:
Thanks for taking the time reviewing the PR and providing awesome suggestions, but I feel the `querySelector` and `querySelectorAll` and using `iterators` looks fine for now. On the other hand, we are not selecting a huge number of nodes from the DOM, so it would not hinder that much performance, like it would just be in milliseconds or so. Also, in the UI sometimes having a little slowness is necessary otherwise the UI might become too snappy, which hinders good user experience like for example, let's take this scenario. Would you prefer having a button which when being pressed gives a result instantaneously which makes it hard to know whether the result did appear (making you have to see for a few seconds to know that it did)?? Or would you prefer a button which has a bit of delay and then outputs results, which makes it easier to notice the output?? I hope you got my point, with that in mind I would like to suggest that we should leave this aside for now :slightly_smiling_face: .

Also, since there are a lot of changes that need to be made, I think a better idea would be to have a new issue for the changes that need to be made. This would make the PR smaller and more focused. What do you say??

public/static/cookies.js
neon-mmd: Thanks for the suggestion, Any valid reason for making this change?? Can you please take some time explaining the reason in brief :slightly_smiling_face: .

public/templates/cookies_tab.html
neon-mmd: Thank you for making this change is that this will make the code a little less readable because the sentence will be very long. Anyone working on this file will have to scroll horizontally to see the whole sentence, which is not good, What do you say?

public/templates/cookies_tab.html
neon-mmd: Thank you for making this change is that this will make the code a little less readable because the sentence will be very long. Anyone working on this file will have to scroll horizontally to see the whole sentence, which is not good, What do you say?

public/templates/engines_tab.html
neon-mmd: Thank you for making this change is that this will make the code a little less readable because the sentence will be very long. Anyone working on this file will have to scroll horizontally to see the whole sentence, which is not good, What do you say?

public/templates/cookies_tab.html
neon-mmd: Thanks for the suggestion, Any valid reason for making this change?? Can you please take some time explaining the reason in brief :slightly_smiling_face: .

neon-mmd:
The change with add links I feel can also be a good first issue I believe, so I would suggest not having the change in this PR. Maybe we could add a new issue for the same and someone could add it with a new PR :slightly_smiling_face: , What do you say?

alamin655 (approved):

alamin655:
> The change with add links I feel can also be a good first issue I believe, so I would suggest not having the change in this PR. Maybe we could add a new issue for the same and someone could add it with a new PR 🙂 , What do you say?

Yes, nice idea.

alamin655:
> Thanks for taking the time reviewing the PR and providing awesome suggestions, but I feel the `querySelector` and `querySelectorAll` and using `iterators` looks fine for now. On the other hand, we are not selecting a huge number of nodes from the DOM, so it would not hinder that much performance, like it would just be in milliseconds or so. Also, in the UI sometimes having a little slowness is necessary otherwise the UI might become too snappy, which hinders good user experience like for example, let's take this scenario. Would you prefer having a button which when being pressed gives a result instantaneously which makes it hard to know whether the result did appear (making you have to see for a few seconds to know that it did)?? Or would you prefer a button which has a bit of delay and then outputs results, which makes it easier to notice the output?? I hope you got my point, with that in mind I would like to suggest that we should leave this aside for now 🙂 .
> 
> Also, since there are a lot of changes that need to be made, I think a better idea would be to have a new issue for the changes that need to be made. This would make the PR smaller and more focused. What do you say??

Ya a new issue would be better.

alamin655:
Thanks!! :slightly_smiling_face:  It's good to go for now, so let's merge it.

PR #146, opened 2023-07-15
✨ Provide a way to allow users to select different engines

### LINKED ISSUE #81: Improve async multithreading for aggregation

### LINKED ISSUE #145: Provide a way to allow users to select different engines
Write handle engine selection code using `match` statement with string based match expressions based on the following scenarios and options:
- [x] configured from the config file with the config options something like this:
``` lua
 upstream_engines = {duckduckgo = true, searx = false}
```
- [x] an error occurs from the upstream search engine, so move to another upstream search engine.
- [x] handle the case that none of the upstream engines provide results.
- [x] the first engine to be selected case like the first engine that is to be selected to initialize `results_map` with.
- [x] Create a common behavior between search engine code using traits.
- [x] #81

neon-mmd:
## What does this PR do?

This PR provide a new feature to allow users to select between different upstream search engines either from the settings page on the website or from the config file. Also, this PR improves the asynchronous multithreading way of fetching results by taking into account that more engines could be added in the future.

This change is essential as this provides a way for the user to select between different search engines, which improves the privacy of the user. Also, the improvement in the asynchronous multithreading takes into account that more engines could be added in the future.

It can be tested by installing and running `Websurfx` as mentioned in the docs and on the `readme` and by launching the browser and thoroughly testing. Selecting the engines from the setting page and saving it, and also testing it from the config file without having any cookie saved from the settings page.

## Author's checklist

- [x] configured from the config file with the config options something like this:
``` lua
 upstream_engines = {duckduckgo = true, searx = false}
```
- [x] an error occurs from the upstream search engine, so move to another upstream search engine.
- [x] handle the case that none of the upstream engines provide results.
- [x] the first engine to be selected case like the first engine that is to be selected to initialize `results_map` with.
- [x] Create a common behavior between search engine code using traits.
- [x] #81 
- [x] Fixing bugs associated while improving the asynchronous multithreading code.
- [x] Add documentation to code 
- [x] Bump the app version

## Related issues

Closes #145 
Closes #81 

## Notes

Special thanks to Sabrina for fixing the bugs associated while improving the asynchronous multithreading code.


src/server/routes.rs
```diff
@@ -104,21 +118,55 @@ pub async fn search(
...
+    query: String,
     page: u32,
+    req: HttpRequest,
 ) -> Result<SearchResults, Box<dyn std::error::Error>> {
     //Initialize redis cache connection struct
     let mut redis_cache = RedisCache::new(config.redis_url.clone())?;
     // fetch the cached results json.
     let cached_results_json = redis_cache.cached_json(&url);
-    // check if fetched results was indeed fetched or it was an error and if so
+    // check if fetched catch results was indeed fetched or it was an error and if so
```
xffxff: typo? catch -> cache?

src/results/aggregation_models.rs
```diff
@@ -124,12 +143,19 @@ impl RawSearchResult {
...
+/// * `empty_result_set` - Stores a boolean which indicates that no engines gave a result for the
+/// given search query.
 #[derive(Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct SearchResults {
     pub results: Vec<SearchResult>,
     pub page_query: String,
     pub style: Style,
+    pub engine_errors_info: Vec<EngineErrorInfo>,
+    pub empty_result_set: bool,
```
xffxff: Why do we need an `empty_result_set`? It can be inferred through the `results`. Keeping both may result in inconsistency, meaning that the `results` are not empty but the `empty_result_set` is true.

src/server/routes.rs
```diff
@@ -104,21 +118,55 @@ pub async fn search(
...
     match cached_results_json {
         Ok(results_json) => Ok(serde_json::from_str::<SearchResults>(&results_json).unwrap()),
         Err(_) => {
-            let mut results_json: crate::results::aggregation_models::SearchResults =
-                aggregate(query, page, config.aggregator.random_delay, config.debug).await?;
+            // check if the cookie value is empty or not if it is empty then use the
+            // default selected upstream search engines from the config file otherwise
+            // parse the non-empty cookie and grab the user selected engines from the
+            // UI and use that.
+            let mut results_json: crate::results::aggregation_models::SearchResults = match req
```
xffxff: `results_json` is not a json. It would be better to remove the suffix.

src/results/aggregator.rs
```diff
@@ -3,22 +3,36 @@
...
+/// by matching over the selected search engines and adding the selected ones to the vector which
+/// is then used to create an async task vector with `tokio::spawn` which returns a future which
+/// is then awaited on in another loop and then all the collected results is filtered for errors
+/// and proper results and if an error is found is then sent to the UI with the engine name and the
+/// error type that caused it by putting them finallt in the returned `SearchResults` struct. Also
+/// the same process also removes duplicate results and if two results are found to be from two or
+/// more engines then puts their names together to show the results are fetched from these upstream
+/// engines and then removes all data from the HashMap and puts into a struct of all results aggregated
+/// into a vector and also adds the query used into the struct this is neccessory because otherwise the
+/// search bar in search remains empty if searched from the query url.
```
xffxff: I've noticed that you don't seem to like using punctuation. I'm not sure if it's intentional. As someone whose English is a second language, I find it difficult to read and understand without proper punctuation :)

src/results/aggregator.rs
```diff
@@ -53,41 +71,85 @@ pub async fn aggregate(
...
+                }
+                Err(error_type) => {
+                    engine_errors_info.push(EngineErrorInfo::new(
+                        error_type.downcast_ref::<EngineError>().unwrap(),
+                        upstream_search_engines[counter].clone(),
+                    ));
+                    counter += 1
+                }
+            }
+        }
```
xffxff, line 178: It would be helpful to provide some explanations about the purpose of this block, `outputs.iter().for_each(...)`. Particularly, it would be good to clarify why we are separating the consideration fi the output is initial.

src/server/routes.rs
neon-mmd: Yes, It is a typo, I am sorry it should have been `cache` instead of `catch` :smile: .

src/results/aggregation_models.rs
neon-mmd: Ok I see, thanks :heart:  for pointing that out I will fix this right away :rocket: :slightly_smiling_face: .

src/server/routes.rs
neon-mmd: Actually my intention for naming it that way was that at the end when the data is returned to the UI for rendering it gets converted into `json` data, but I think I will go with your suggestion and fix it :slightly_smiling_face:.

src/results/aggregator.rs
neon-mmd: I am sorry, I actually don't have any specific issues with punctuation, just that I have a habit of forgetting to add it :smile: . But I am already in the process of changing my habit, and also I will fix this as well :slightly_smiling_face: . Thanks for pointing this out :heart: .

src/results/aggregator.rs
neon-mmd: Thanks :heart: for pointing this, I will fix this right away :rocket: :slightly_smiling_face: .

neon-mmd:
xffxff, as per your suggestions, I have added all the fixes and additions to the code. I hope everything looks good now :slightly_smiling_face: .

xffxff:
@neon-mmd  Everything looks good to me now. Your code comments are exceptionally clear and well-written.

xffxff (approved):

neon-mmd:
As per your review, since everything looked good, henceforth I have merged this PR :slightly_smiling_face:. Thanks :heart: for taking the time to review this PR. I really appreciate it :+1: :slightly_smiling_face: .

PR #165, opened 2023-07-30
🛠️ Handle page value returned from search query parameter

### LINKED ISSUE #164: 🐛 `Searx` engine does not return any results on the first page
**What happened?**
The `Searx` engine does not provide any results on the first page when queried something like this `http://127.0.0.1:8080/search?q=ls`. 

**Expected behavior**
The results shown on the first page should include the `Searx` results.

**Screenshots**
image

**Additional context**
The problem may seem to be lying within the logic associated with page handling in `searx.rs` when the default page is passed from `routes.rs` code to `searx.rs` code.

neon-mmd:
## What does this PR do?

This PR fixes #164 by adds and improves code to handle the page value passed from the search query parameter.

This change is essential as it fixes the issue of having no search results from Searx engine on the first page when queried with `http://127.0.0.1:8080/search?q=ls`.

It can be tested by installing and running `Websurfx` as mentioned in the `docs` and on the `readme` and by launching the browser and thoroughly testing. Searching the web with query parameters and without query parameters and providing different values to it.

## Author's checklist

- [x] add code to handle page value
- [x] improve code to handle page value by changing the match patterns
- [x] bump the app version to `v0.15.2` 

## Related issues

Closes #164

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -83,12 +83,12 @@ dependencies = [
 
 [[package]]
 name = "actix-macros"
-version = "0.2.3"
+version = "0.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "465a6172cf69b960917811022d8f29bc0b7fa1398bc4f78b3c466673db1213b6"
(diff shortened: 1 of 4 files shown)


xffxff (dismissed):

alamin655 (approved):

PR #171, opened 2023-08-02
✨ Config option to allow users to customize the number of threads used by the app

### LINKED ISSUE #169: ✨ Config option to allow users to customize the number of threads used by the app.
Provide a new config option to allow users to customize the amount of threads to be given to the app. 

The new config option will look something like this:

```lua
threads = 10  -- the amount of threads that the app will use to run (the value should be greater than 0).
```

neon-mmd:
## What does this PR do?

This PR provides a new config option `threads` and improves the logs generated by the app.

This PR is essential as it allows the user to control the amount of threads that the app should use according to their system requirement. Also, this PR improves the logging, which is essential as this allows errors to also be logged from the config parser module `parser.rs` which were previously not being logged.

It can be tested by installing and running `Websurfx` as mentioned in the `docs` and on the `readme` and by launching the browser and thoroughly testing. Checking the logs that get displayed on standard output when either `debug` or `logging` is set to true or false.

## Author's checklist

- [x] adds a new config option `threads`.
- [x] improves the logging functionality in the app.
- [x] bumps the app version to `v0.16.0`. 

## Related issues

Closes #169

### DIFF
--- a/src/config/parser.rs
+++ b/src/config/parser.rs
@@ -54,12 +54,17 @@ impl Config {
     /// A function which parses the config.lua file and puts all the parsed options in the newly
     /// constructed Config struct and returns it.
     ///
+    /// # Arguments
+    ///
+    /// * `logging_initialized` - It takes a boolean which ensures that the logging doesn't get
+    /// initialized twice.
+    ///
     /// # Error
     ///
     /// Returns a lua parse error if parsing of the config.lua file fails or has a syntax error
     /// or io error if the config.lua file doesn't exists otherwise it returns a newly constructed
     /// Config struct with all the parsed config options from the parsed config file.
-    pub fn parse() -> Result<Self, Box<dyn std::error::Error>> {
+    pub fn parse(logging_initialized: bool) -> Result<Self, Box<dyn std::error::Error>> {
         Lua::new().context(|context| -> Result<Self, Box<dyn std::error::Error>> {
             let globals = context.globals();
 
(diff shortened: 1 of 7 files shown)


src/config/parser.rs
```diff
@@ -51,19 +54,56 @@ impl Config {
...
+                    log_level = LevelFilter::Info;
+                } else if debug {
+                    log_level = LevelFilter::Trace;
+                };
+                env_logger::Builder::new().filter(None, log_level).init();
+            }
+
+            let threads: u8 = if parsed_threads == 0 {
+                    let total_num_of_threads:usize =  available_parallelism()?.get() /2;
+                if debug || logging {
```
xffxff, line 94: Do we have to check the `debug` and `logging` flag every time we want to add a log? It feels quite tedious. In my opinion, the `debug` and `logging` flags should only affect the logging level. We should be able to log in anywhere, considering only the log level we want to choose. @neon-mmd What do you think

src/config/parser.rs
```diff
@@ -51,19 +54,56 @@ impl Config {
...
+            // Check whether logging has not been initialized before.
+            if logging_initialized {
+                // Initializing logging middleware with level set to default or info.
+                let mut log_level: LevelFilter = LevelFilter::Off;
+                if logging && debug == false {
+                    log_level = LevelFilter::Info;
+                } else if debug {
+                    log_level = LevelFilter::Trace;
+                };
+                env_logger::Builder::new().filter(None, log_level).init();
```
xffxff, line 89: We initialize the logger in a function called `parse`,  I did not expect this function to have such a side effect. However, considering that it is dependent on the parsed configuration and we also want to add some logs in the `parse` function, I can not think of a better place to do this :joy:

src/config/parser.rs
neon-mmd: Yes, I think you are right, maybe we should make it log level based. But I think we should not do it in this PR because this PR will become huge as there are a lot of this checks in the program, and we will need to replace them all :slightly_smiling_face: . So I think let's merge this PR and work this suggestion in the new PR. What do you say ??

Also, I would like to have some suggestions on this. I think a `trace` log level is great, but after doing some research, I found that this log level should only be used when tracing third party libraries or code. So, I think `trace` the log level should be only when anyone builds the app with dev environment by allowing the `Cargo.toml` to set an `enviroment` variable as `dev`. This will then allow the program to read this variable and then turn on tracing. Otherwise, in production, there will only two logging level `debug` and `info`. What do you say ?? :slightly_smiling_face:

Here are some links on my research:
- https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels
- https://sematext.com/blog/logging-levels/ 
- https://betterstack.com/community/guides/logging/log-levels-explained/

src/config/parser.rs
neon-mmd: It is not a side effect though :joy: , but actually in the main function previously when logging was being initialized it was being initialized after the config was being parsed, which caused any logging information in the parse function getting ignored. Also, now since initialization of the logging depends on the `debug` and `logging` options, so I had to move it to the parse function. I tried other solutions, but they didn't work, so I think this is the only solution I guess :smile: .

src/config/parser.rs
xffxff: > This will then allow the program to read this variable and then turn on tracing. Otherwise, in production, there will only two logging level `debug` and `info`. What do you say ?? 🙂

Yes, that makes sense. 

Additionally, thanks for providing the links, I have learned a lot from them

xffxff (approved):

PR #176, opened 2023-08-03
🔧 Config options to change the logging level with improved logging

### LINKED ISSUE #170: 🔧 Add more detailed logging for debugging
I'm creating this issue to request an enhancement to the current logging system of websurfx. While searching on the web, I encountered a situation where I received an empty result in the web interface. However, the current server logs did not provide any information to understand the reason behind this.

image
image

It was unclear whether the empty result was due to a request timeout, an issue with the upstream engine, or some other unknown factor. More detailed logs would greatly help in troubleshooting these kinds of issues in the future.

### LINKED ISSUE #173: 🔧 Config options to change the logging level with improved logging
Remove unnecessary `debug` and `logging` level checks to do logging when either of the two config options `debug` and `logging` are set to true or false. Also, provide a mechanism to only set `Trace` log level when the program is either or build in `development` environment.

neon-mmd:
## What does this PR do?

This PR improves the logging based on levels (This is based on the suggestion provided by xffxff in PR #171 and based on issue #170).

This PR is essential as it removes the unnecessary checks using `logging` and `debug` config options and environment variable to restrict logs. Also, it improves logging by adding logging for `engine errors` as well. 

It can be tested by installing and running `Websurfx` as mentioned in the `docs` and on the `readme` and by launching the browser and thoroughly testing. By checking logs by providing environment variables and setting either `logging` or `debug option to true or false`.

 For example, using an environment variable:

``` shell
PKG_ENV=dev ./target/debug/websurfx
```

## Author's checklist

- [x] improve the logging based on config option and environment variable.
- [x] add logging for `engine errors`.
- [x] bump the app version to `v0.16.2` 

## Related issues

Closes #173 
Closes #170 

## Notes

This PR should only be merged after PR #171 gets merged. As this PR depends on the PR #171 .

### DIFF
--- a/src/config/parser.rs
+++ b/src/config/parser.rs
@@ -4,7 +4,7 @@
 use super::parser_models::Style;
 use log::LevelFilter;
 use rlua::Lua;
-use std::{collections::HashMap, format, fs, io::Write, path::Path, thread::available_parallelism};
+use std::{collections::HashMap, format, fs, path::Path, thread::available_parallelism};
 
 // ------- Constants --------
 static COMMON_DIRECTORY_NAME: &str = "websurfx";
@@ -79,26 +79,26 @@ impl Config {
 
             // Check whether logging has not been initialized before.
             if logging_initialized {
-                // Initializing logging middleware with level set to default or info.
-                let mut log_level: LevelFilter = LevelFilter::Off;
-                if logging && debug == false {
-                    log_level = LevelFilter::Info;
-                } else if debug {
-                    log_level = LevelFilter::Trace;
-                };
-                env_logger::Builder::new().filter(None, log_level).init();
+                if let Ok(pkg_env_var) = std::env::var("PKG_ENV"){
+                    if pkg_env_var.to_lowercase() == "dev" {
+                        env_logger::Builder::new().filter(None, LevelFilter::Trace).init();
+                    }
+                } else {
+                    // Initializing logging middleware with level set to default or info.
+                    let mut log_level: LevelFilter = LevelFilter::Error;
+                    if logging && debug == false {
+                        log_level = LevelFilter::Info;
+                    } else if debug {
+                        log_level = LevelFilter::Debug;
+                    };
+                    env_logger::Builder::new().filter(None, log_level).init();
+                }
             }
 
             let threads: u8 = if parsed_threads == 0 {
-                    let total_num_of_threads:usize =  available_parallelism()?.get() /2;
-                if debug || logging {
-                    log::error!("Config Error: The value of `threads` option should be a non zero positive integer");
-                    log::info!("Falling back to using {} threads", total_num_of_threads)
-                } else {
-                    std::io::stdout()
-                        .lock()
-                        .write_all(&format!("Config Error: The value of `threads` option should be a non zero positive integer\nFalling back to using {} threads\n", total_num_of_threads).into_bytes())?;
-                };
+                let total_num_of_threads: usize =  available_parallelism()?.get() / 2;
+                log::error!("Config Error: The value of `threads` option should be a non zero positive integer");
+                log::info!("Falling back to using {} threads", total_num_of_threads);
                 total_num_of_threads as u8
             } else {
                 parsed_threads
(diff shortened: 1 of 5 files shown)


src/config/parser.rs
```diff
@@ -79,26 +79,26 @@ impl Config {
...
+                        env_logger::Builder::new().filter(None, LevelFilter::Trace).init();
+                    }
+                } else {
+                    // Initializing logging middleware with level set to default or info.
+                    let mut log_level: LevelFilter = LevelFilter::Error;
+                    if logging && debug == false {
+                        log_level = LevelFilter::Info;
+                    } else if debug {
+                        log_level = LevelFilter::Debug;
+                    };
```
xffxff, line 93: I have an idea that I'd like to discuss with you. Instead of using a boolean configuration option like `logging=true/false` to enable or disable logging (which currently only affects the logging level, as logs above INFO are still recorded when `logging=false`). How about supporting the direct setting of the logging level? For example, instead of using `logging=false`, we could use `logging=info`, or `logging=debug`

neon-mmd: I think this is a good idea. I don't know why I didn't think about this before :joy: but there is one thing we will need to rename the config option from `logging` to `logging` so that it does not add confusion for the user. Also, by renaming it to `logging-mode` I feel it will become more sensible. Moreover, instead of calling the options info, debug I think we can give a bit of easier options to remember for the user like `on`, `verbose` or `off`. What do you say ? :slightly_smiling_face:

xffxff: I believe using the term `verbose` would be more suitable than `on` or `off`, since the logs are not truly disabled even when we set it to `off` :)

neon-mmd: Yes I agree, but I intend to have `off` as a value because it disables all the other necessary information like `info`, `debug`, etc. log levels except the `error` level. This is necessary because the user should know what went wrong when they started the app, like if they had put a wrong value in config or anything. I hope you got my point :slightly_smiling_face:

xffxff: I see. If the `off` option disables the logging with `info` level, it make sense to me because there could be many logs with that level

Should we change it to `logging-mode` in the next PR, or do it in this PR?

neon-mmd: I think let's do it in this PR because the change will be related. Also, it won't be a big change, so let's do it in this PR itself :slightly_smiling_face: .

neon-mmd: After trying to implement the solution as you had suggested here. It seems that there is a problem when eliminating debug option, as it used to check to disable random delays between requests when the app is put in debug mode for testing purposes. So, I think if we eliminate the `debug` option, we might break some parts of the app. So I think it would be better not to implement this solution :slightly_smiling_face: . What do you say?

xffxff: I don't think we should delete the `debug` option, my opinion is that we don't need to check if `debug` and `logging` is true every time we add logs.

Additionally, why do we have both `debug` and `production_use` options? Do we only need one of them?

neon-mmd: > I don't think we should delete the `debug` option, my opinion is that we don't need to check if `debug` and `logging` is true every time we add logs.
> 
> Additionally, why do we have both `debug` and `production_use` options? Do we only need one of them?

The reason to provide the `debug` option is to have some things in the app only be set for debugging way when this option is enabled like disabling `random delays`, etc and many other things in the future. While `production_mode` option allows the user to enable `random_delays` or any other thing which is good for server use. Now coming back to your question It is valid but let's say if we only debug mode then disabling it would enable `random delays` but if the user doesn't want `random delays` because it is just for personal use then the user will need to enable it but then enabling it means it will also enable other things which is not really intended for the use. Also it might add up for a lot of confusion so I think having both these options is correct otherwise it will be a really bad experience for the user. What do you say ? :slightly_smiling_face:

xffxff: Currently, both `debug` and `logging` options affect the log level, and `debug` and `production_use` options affect `random_delay`. The different combinations of their values result in different settings, which is confusing for me and leads to some unclear configurations. I suggest that each option in `config.lua` should independently control specific settings, without multiple options affecting a single setting simultaneously.

However, let's merge this PR first and create a new issue to discuss this problem

xffxff (approved):

PR #177, opened 2023-08-05
📜 Revise the `docs` to remain in sync with the current changes

### LINKED ISSUE #175: 📜 Revise the `docs` to remain in sync with the current changes
Revise the `docs` to add documentation covering the new option and feature provided by the PR https://github.com/neon-mmd/websurfx/pull/171 .

neon-mmd:
## What does this PR do?

This PR adds explanation for the new config option proposed in the PR #171. 

This PR is essential as it keeps the `docs` in sync with the current changes that were submitted in the PR #171 .

## Author's checklist

- [x] Add explanation for the new config option `threads`.
- [x] bump the app version to `v0.16.3` 

## Related issues

Closes #175

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -23,6 +23,7 @@ Some of the configuration options provided in the file are stated below. These a
 
 - **logging:** An option to enable or disable logs.
 - **debug:** An option to enable or disable debug mode.
+- **threads:** The amount of threads that the app will use to run (the value should be greater than 0).
 
 ## Server
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2684,18 +2684,18 @@ checksum = "388a1df253eca08550bef6c72392cfe7c30914bf41df5269b68cbd6ff8f570a3"
 
 [[package]]
 name = "serde"
-version = "1.0.180"
+version = "1.0.181"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0ea67f183f058fe88a4e3ec6e2788e003840893b91bac4559cabedd00863b3ed"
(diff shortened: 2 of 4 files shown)


docs/configuration.md
```diff
@@ -23,6 +23,7 @@ Some of the configuration options provided in the file are stated below. These a
 
 - **logging:** An option to enable or disable logs.
 - **debug:** An option to enable or disable debug mode.
+- **threads:** The amount of threads that the app will use to run (the value should be greater than 0).
```
xffxff, line 26: Should we document to users that if the `threads` option is set to 0, we will automatically set it to `thread::available_parallelism()/2`?

alamin655: Well, it does say that the value should be greater than 0, but we can add a note about it.

neon-mmd: > Should we document to users that if the `threads` option is set to 0, we will automatically set it to `thread::available_parallelism()/2`?

I think we might not need to document that if we put it 0 part because it is just an error handling thing which is used to handle the 0 case value. Like if somebody does set it to 0 by mistake then it will handle the error by falling back to 0. :slightly_smiling_face: So I don't think we need to document that part. What do you say ?

alamin655: > > Should we document to users that if the `threads` option is set to 0, we will automatically set it to `thread::available_parallelism()/2`?
> 
> I think we might not need to document that if we put it 0 part because it is just an error handling thing which is used to handle the 0 case value. Like if somebody does set it to 0 by mistake then it will handle the error by falling back to 0. 🙂 So I don't think we need to document that part. What do you say ?

I agree with you.

alamin655 (approved):

PR #184, opened 2023-08-06
🔧 Cache next and previous pages on search

### LINKED ISSUE #160: 🔧 Cache next and previous pages on search
## Why?
When searching, the backend only requests the current page, but it should also get the result for the next page and previous page. This would avoid extra loading time when going to the next page or previous page, as results are already cached. 

### How?
We can simply create an async task that is run after the requests are done, which will fetch the results, cache them, so that they can be served as requested without delay.

neon-mmd:
## What does this PR do?

This PR adds code to simultaneously/concurrently cache the previous and next pages on searching the current page.

This change is essential as it removes the extra loading time when going to previous or next pages as the results are already cached beforehand.

It can be tested by installing and running `Websurfx` as mentioned in the `docs` and on the `readme` and by launching the browser and thoroughly testing. By searching the current page and checking the Redis server stored keys using the `redis-cli`. 

```shell
redis-cli -p 8082
127.0.0.1:8082> KEYS *
```

## Author's checklist

- [x] add code to improve load times by caching the previous & next pages
- [x] bump the app version to `v0.16.5` 

## Related issues

Closes #160

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1817,9 +1817,9 @@ checksum = "0ab1bc2a289d34bd04a330323ac98a1b4bc82c9d9fcb1e66b63caa84da26b575"
 
 [[package]]
 name = "openssl"
-version = "0.10.55"
+version = "0.10.56"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


alamin655 (approved):

PR #190, opened 2023-08-22
✨ Filter user specified content from search results using a `blacklist` and `Allowlist`.

### LINKED ISSUE #163: ✨ Filter user specified content from search results using a `blacklist` and `Allowlist`.
Provide a way to filter out all user specified content to be filtered out using a blacklist and allowlist by checking the result header, URL, and result description using regex.

For example:

if the result shows:

``` markdown
# Some nsfw header
nsfw-url
nsfw desc
```

Then filter out this result by matching the NSFW word or URL by checking first the header then URL and finally the description if any of the three contains the NSFW word when matching it against the `blacklist`. Then filter out the result out by removing it from the `result_map` and adding in a temporary `blacklist_map`. Also, after the results are filtered, check whether the filtered out results do not contain false positives by checking each result in the `blacklist_map`. This is similar to how it was done previously by using regex but this time against the `allowlist`. If a false positive result is found, then remove it from the `blacklist_map` and insert it back into the `result_map`.

neon-mmd:
## What does this PR do?

This PR adds the ability to filter out aggregated results based on lists. 

This change is essential as this provides more control to the user on what results they want to see and what not.

It can be tested by installing and running `Websurfx` as mentioned in the `docs` and on the `readme` and by launching the browser and thoroughly testing. By setting a regex in the block-list/allow-list file and searching the blocked or allowed content and seeing whether it gets blocked or allowed in the search results. 

## Author's checklist

- [x] add code to filter search lists based on lists.
- [x] reimplement config file/ theme file/ list files path code to avoid code duplication.
- [x] bump the app version to `v0.18.0`   

## Related issues

Closes #163


src/results/aggregator.rs
```diff
@@ -143,11 +151,46 @@ pub async fn aggregate(
...
+    let results: Vec<SearchResult> = result_map.into_values().collect();
 
     Ok(SearchResults::new(
         results,
         query.to_string(),
         engine_errors_info,
     ))
 }
+
+fn filter_with_lists(
```
xffxff, line 178: I think this implementation should be correct, but it would be better if we add some unit tests for this function

neon-mmd: I think we could add all the tests at the end once we are done with everything by putting a separate PR for this. I think this would be better. What do you say ? :slightly_smiling_face: .

Also, to be honest, I am not really the right person for writing tests because I am not good at it :joy: . So I think maybe we might need someone who has a good amount of experience to write tests. What do you say ?

xffxff: > I think we could add all the tests at the end once we are done with everything by putting a separate PR for this. I think this would be better. What do you say ? 🙂 .

That's fine. Let's go ahead and do it that way for now

> Also, to be honest, I am not really the right person for writing tests because I am not good at it 😂 . So I think maybe we might need someone who has a good amount of experience to write tests. What do you say ?

I can help write some unit tests

xffxff (approved):

Issue #179, opened 2023-08-05
🔧 More unit tests, integration test, and doc tests


neon-mmd:
Add more unit tests, integration tests and doc tests to allow better maintainability and testing of the code.

neon-mmd:
@xffxff are you interested in working on this issue? If you are, then I will assign this issue to you 🙂 .

xffxff:
> @xffxff are you interested in working on this issue? If you are, then I will assign this issue to you 🙂 .

I'm happy to help with this, but there's no need to assign this issue specifically to me. I believe that improving tests is a never-ending task that requires the contributions of all of us. Everyone is welcome to work on this, and this issue could also be a good first issue for those who wish to begin contributing

willser:
I would like to add some ut for understanding this repo.Do you have any suggestion such as where I should start.

neon-mmd:
> I would like to add some ut for understanding this repo.Do you have any suggestion such as where I should start.

Yes sure :slightly_smiling_face:, I think first I would suggest you to go through the [developer](https://github.com/neon-mmd/websurfx/blob/rolling/docs/developing.md) docs to better understand the structure of the project. Then I would recommend you to start writing tests first for covering all the routes in project like for search route, etc. to ensure that all the routes are returning data as expected. Then you can explore other functions which could benefit from writing tests for it. :slightly_smiling_face: 

Also, since you want to work on this issue, so I am assigning this issue to you. You may start working on it right away :rocket: :slightly_smiling_face: .

neon-mmd:
Willser, it has been over a month, any progress on this issue so far? We would like to know :slightly_smiling_face: .

neon-mmd:
@willser, it has been over a month, any progress on this issue so far? We would like to know :slightly_smiling_face: .

willser:
> any progress on this issue so far?

@neon-mmd I am so sorry,I am very busy for my work recently,I am not sure what time I can start this issue.Sorry. 😢

neon-mmd:
> > any progress on this issue so far?
> 
> @neon-mmd I am so sorry,I am very busy for my work recently,I am not sure what time I can start this issue.Sorry. 😢

Sorry for the late reply.
Ok sure, no problem take your time but we would suggest keeping us updated on this issue. Also, if you are unable to work on this issue at all, then we would suggest instead maybe un-assigning yourself from the issue and you can then explore some other quick to work on issue in the project. :slightly_smiling_face:

PR #191, opened 2023-08-24
🧹 Unit tests for the function `filter_with_lists`

xffxff:
## What does this PR do?

This PR adds the documentation and tests for the function `filter_with_lists`. Also, this PR adds progress on the issue #179.

This change is essential as this provides the developer with the ability to test the function more extensively.

It can be tested by cloning the repository and running `cargo test` command.

``` shell
git clone https://github.com/neon-mmd/websurfx
cd websurfx
cargo test
```

## Author's checklist

- [x] add docs for `filter_with_lists`
- [x] some unit tests for `filter_with_lists`
- [x] bump the app version to `v0.18.1`   

## Related issues

#179

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ once_cell = {version="1.18.0"}
 error-stack = {version="0.3.1"}
 async-trait = {version="0.1.73"}
 regex = {version="1.9.3", features=["perf"]}
+tempfile = "3.8.0"
 
 [dev-dependencies]
 rusty-hook = "^0.11.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "websurfx"
-version = "0.18.0"
+version = "0.18.1"
 edition = "2021"
 description = "An open-source alternative to Searx that provides clean, ad-free, and organic results with incredible speed while keeping privacy and security in mind."
 repository = "https://github.com/neon-mmd/websurfx"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -27,6 +27,7 @@ once_cell = {version="1.18.0"}
 error-stack = {version="0.3.1"}
 async-trait = {version="0.1.73"}
 regex = {version="1.9.3", features=["perf"]}
+tempfile = "3.8.0"
 
 [dev-dependencies]
 rusty-hook = "^0.11.2"
```
```suggestion

[dev-dependencies]
rusty-hook = "^0.11.2"
tempfile = "3.8.0"
```

Thanks :heart: for working on this PR, I really appreciate it :+1:  but I would suggest you to put the `tempfile` dependency under `dev.dependencies` as we don't want to include in the main binary file and only have it included when we run `cargo tests`.

For more information on `dev.dependencies` I would suggest going through these docs here: 
- https://doc.rust-lang.org/rust-by-example/testing/dev_dependencies.html#development-dependencies
- https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies

neon-mmd (changes requested):

Cargo.toml
xffxff: I agree. Thank you for pointing that out!

neon-mmd (approved):
Thanks :heart: once again for working on this. Since everything looks good, I will merge this PR right away.

PR #194, opened 2023-08-30
⚙️ Optimize and make code more idiomatic (part - I)

### LINKED ISSUE #178: ⚙️ Make code more idiomatic
Make code more idiomatic whenever essential, like changing loops to iterators and if statements to match statements.

### LINKED ISSUE #180: 🔧 Optimize code for speed and memory usage
Optimize code to improve the performance and reducing the memory usage of the app by taking advantage of clever tricks and techniques.

neon-mmd:
## What does this PR do?

This PR adds several optimizations to the code to reduce the memory usage by using clever techniques like using `smallvec` crate for vectors with small data size and changing function signatures to reduce memory allocations. Also, this PR adds pool implementation for Redis connections to improve performance. Moreover, this PR adds countless other optimizations to improve speed. This PR also finishes issues #180 and #178

This change is essential because this improves the speed, efficiency and reduces memory usage which aligns with the goal of the project.

It can be tested by installing and running `Websurfx` as mentioned in the `docs` and on the `readme` and by launching the browser and thoroughly testing. By testing against the `dhat` profiler and various other profilers. 

## Author's checklist

- [x] change function signatures to reduce memory allocations
- [x] change vectors to `smallvecs` to reduce memory allocation for smaller data sizes.
- [x] replace `once_cell` with `oncelock` in std library to improve speed.
- [x] add `dhat` profiler for testing memory usage.  
- [x] change `to_string()` to `to_owned()` to improve speed.
- [x] replace `rlua` with `mlua` for speed.
- [x] replace `jemalloc` with `mimalloc` for speed.
- [x] add compress middleware to compress headers to reduce memory usage. 
- [x] change Redis cache implementation slightly to initialize the struct only once.  
- [x] bump the app version to `v0.18.4`  

## Additional Resources
- How to optimize rust code:  https://youtu.be/nO-BFKyWYgE?feature=shared
- How to use `dhat` profiler:  https://www.youtube.com/watch?v=AJhKaoyc4pY

## Related issues

Closes #180 
Closes #178

## Additional Notes

Some checks may not pass as they will fail because of the `mlua` crate which depends on the `luajit` binary, which requires it to be available on the system beforehand. So the CI jobs would require to be fixed in order for them to work as usual. Which I think could be fixed in this release. What do you say?

### DIFF
--- a/src/bin/websurfx.rs
+++ b/src/bin/websurfx.rs
@@ -6,7 +6,7 @@
 use std::net::TcpListener;
 use websurfx::{config::parser::Config, run};
 
-/// A dhat heap memory profiler 
+/// A dhat heap memory profiler
 #[cfg(feature = "dhat-heap")]
 #[global_allocator]
 static ALLOC: dhat::Alloc = dhat::Alloc;
@@ -22,7 +22,7 @@ async fn main() -> std::io::Result<()> {
     // A dhat heap profiler initialization.
     #[cfg(feature = "dhat-heap")]
     let _profiler = dhat::Profiler::new_heap();
-    
+
     // Initialize the parsed config file.
     let config = Config::parse(false).unwrap();
 
--- a/src/bin/websurfx.rs
+++ b/src/bin/websurfx.rs
@@ -6,6 +6,11 @@
 use std::net::TcpListener;
 use websurfx::{config::parser::Config, run};
 
+/// A dhat heap memory profiler 
+#[cfg(feature = "dhat-heap")]
+#[global_allocator]
+static ALLOC: dhat::Alloc = dhat::Alloc;
+
 /// The function that launches the main server and registers all the routes of the website.
 ///
 /// # Error
@@ -14,6 +19,10 @@ use websurfx::{config::parser::Config, run};
 /// available for being used for other applications.
 #[actix_web::main]
 async fn main() -> std::io::Result<()> {
+    // A dhat heap profiler initialization.
+    #[cfg(feature = "dhat-heap")]
+    let _profiler = dhat::Profiler::new_heap();
+    
     // Initialize the parsed config file.
     let config = Config::parse(false).unwrap();
 
--- a/src/cache/cacher.rs
+++ b/src/cache/cacher.rs
@@ -1,17 +1,26 @@
 //! This module provides the functionality to cache the aggregated results fetched and aggregated
 //! from the upstream search engines in a json format.
 
+use error_stack::Report;
+use futures::future::try_join_all;
 use md5::compute;
-use redis::{Client, Commands, Connection};
+use redis::{aio::ConnectionManager, AsyncCommands, Client, RedisError};
+
+use super::error::PoolError;
 
 /// A named struct which stores the redis Connection url address to which the client will
 /// connect to.
 ///
 /// # Fields
 ///
-/// * `redis_connection_url` - It stores the redis Connection url address.
+/// * `connection_pool` - It stores a pool of connections ready to be used.
+/// * `pool_size` - It stores the size of the connection pool (in other words the number of
+/// connections that should be stored in the pool).
+/// * `current_connection` - It stores the index of which connection is being used at the moment.
 pub struct RedisCache {
-    connection: Connection,
+    connection_pool: Vec<ConnectionManager>,
+    pool_size: u8,
+    current_connection: u8,
 }
 
 impl RedisCache {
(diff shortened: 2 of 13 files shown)


src/cache/cacher.rs
```diff
@@ -1,29 +1,52 @@
...
+
+        for _ in 0..pool_size {
+            tasks.push(client.get_tokio_connection_manager());
+        }
+
+        let redis_cache = RedisCache {
+            connection_pool: try_join_all(tasks).await?,
+            pool_size,
+            current_connection: Default::default(),
+        };
```
xffxff, line 49: Why do we opt to manually create a connection pool? Usually, we would employ a connection pool library such as [`bb8`](https://docs.rs/bb8-redis/latest/bb8_redis/) or [`r2d2`](https://github.com/sorccu/r2d2-redis) to manage connection pooling, which tends to be a more common approach.

(To be honest, I haven't had much experience with Redis, so all of the opinions are based on some internet research:))

src/cache/cacher.rs
```diff
@@ -41,9 +64,42 @@ impl RedisCache {
     /// # Arguments
     ///
     /// * `url` - It takes an url as a string.
-    pub fn cached_json(&mut self, url: &str) -> Result<String, Box<dyn std::error::Error>> {
-        let hashed_url_string = Self::hash_url(url);
-        Ok(self.connection.get(hashed_url_string)?)
+    pub async fn cached_json(&mut self, url: &str) -> Result<String, Report<PoolError>> {
+        self.current_connection = Default::default();
```
xffxff, line 68: this will always reset `current_connection` to 0, this means the same connection will always be used, and others will be ignored

src/cache/cacher.rs
```diff
@@ -41,9 +64,42 @@ impl RedisCache {
...
+                        result = self.connection_pool[self.current_connection as usize]
+                            .get(hashed_url_string)
+                            .await;
+                        continue;
+                    }
+                    false => return Err(Report::new(PoolError::RedisError(error))),
+                },
+                Ok(res) => return Ok(res),
+            }
+        }
```
xffxff, line 102: If we use a connection pool library, this function will be simpler, maybe like
```rust
let mut conn = self.connection_pool.get().await?;
let result: String = conn.get(url).await?;
Ok(result)
```

src/bin/websurfx.rs
```diff
@@ -6,6 +6,11 @@
 use std::net::TcpListener;
 use websurfx::{config::parser::Config, run};
 
+/// A dhat heap memory profiler
+#[cfg(feature = "dhat-heap")]
+#[global_allocator]
+static ALLOC: dhat::Alloc = dhat::Alloc;
```
xffxff, line 12: It seems that we haven't added `dhat` to the dependencies, but that's okay for this PR. I noticed that we did that in another part of this PR series.

src/cache/cacher.rs
neon-mmd: Yes, I totally agree with you, I would also prefer the same, but it seems that none of the pooling libraries like redis-rs, bb8 or r2d2 provide all the 3 things together which is asynchronous connections, connection-manager and pooling. That's why I had to choose this method and had to manage pooling on our own. Also, I want the 3 features together because I want to provide a very low latency and performant system, that's why I am wanting to have the 3 things together. :slightly_smiling_face:

src/cache/cacher.rs
neon-mmd: I think to understand why I did this, I think I will first explain in brief what the code does. So the code actually what it does is that when a connection from the pool let's call this A fails due to `connection drop error` then this is replaced with another connection B from the pool. Now, Since I am using the connection manager feature, connection A is automatically reconnected in the background, which can take sometime. Which means that if that connection A were to be used, then we would have to wait for it to be re-connected. Which means it will add a time delay (or latency) in some cases it would be a significant delay which is bad you know. Also, there is another scenario where you might even have a situation where the connection drops repeatedly because of some unknown reason. Maybe the cloud service provider who provided the Redis service has some issue or maybe due internet connection lose. That's why what the code does is it replaces A with B. Now what I have done is that when you search the next time then connection A would have been re-connected which means we could re-use it. So that's why I have added the line which you have pointed above. :slightly_smiling_face:

src/cache/cacher.rs
neon-mmd: Yes, you are right I totally agree with you but as I discussed above none of the crates so far provide the 3 features together so that's why I have gone with this method. :slightly_smiling_face:

src/bin/websurfx.rs
neon-mmd: Yes, my bad :sweat_smile:  but I think when I split the PRs I just missed it, but I think let's have that be it in the next PR. :slightly_smiling_face: What do you say?

src/cache/cacher.rs
xffxff: I don't think this implementation of the connection pool is correct. The current implementation doesn't seem to be thread-safe. If multiple threads call `cached_json` simultaneously, they might interfere with each other's usage of the `current_connection` index and the `connection_pool`.

In a typical connection pool, connections are reused as much as possible to avoid the overhead of establishing a new connection. However, in the code provided at https://github.com/neon-mmd/websurfx/blob/a799af312525cb198fd744826ca4118f26bef3c8/src/server/routes.rs#L162, we create a new `RedisCache` for each query.

> Also, I want the 3 features together because I want to provide a very low latency and performant system, that's why I am wanting to have the 3 things together

I don't think the current implementation is faster than `r2d2` or `bb8`. Furthermore, I don't think our system would benefit significantly from optimizing this aspect. Therefore, I suggest conducting benchmarking before prioritizing optimization efforts

neon-mmd: Yes, you are right once again :joy: Actually I have worked on this issue by fixing it in the next part of the series of PRs, so I would also suggest taking a look at it as well :slightly_smiling_face: .

alamin655 (approved):
#

PR #202, opened 2023-09-02
✨ `Safe search` levels and `filter lists` for the app

### LINKED ISSUE #201: ✨ `Safe search` levels and filter lists for the app
Implement the following safe search and filtering features in the app:
- [x] #200 
- [x] #126 
- [x] #199 
- [x] #198

neon-mmd:
## What does this PR do?

This PR provides a new config option `safe search` and a new search parameter with the same name with different levels of sensitive content filtering ranging from 1-4 where 1 and 4 are included.

This change is essential as it allows the user to choose from different safe search levels, which allows the user to opt off sensitive content at varying levels. Options 0, 1 and 2 provide no safe search to moderate safe search while 3 and 4 provide high and aggressive level of safe search which uses filter lists to achieve this which can be configured by the user.

> **Note**
> 1. Options 3 and 4 are the same under the hood but they are slightly different in that. That option 3 does not disallow user from searching sensitive content but filters sensitive content via filter lists and if all the results are filtered then provides a filtered error message to the user wherease the option 4 disallows the user from searching sensitive content via filter lists and displays a dissallowed error message. This has been done so that if any user doesn't want to provide a harsh disallow message to someone when search such content but still want to filter content then option 3 is provided while someone lets a parent doesn't want their child to even search sensitive content then option 4 covers that.
> 2. Also Options 3 and 4 can only be configured via the config file so that the user does not override it for safety reasons. For example the scenario discussed above the parent and the child.
> 3. Allowlist has also been added with the blocklist so as to prevent the side effects caused by blocklisting a particular word or site via regex.
> 4. The safe search url parameter only allows going from 1 to 2. This has been done for safety reasons as mentioned previously.

It can be tested by installing and running `Websurfx` as mentioned in the `docs` and on the `readme` and by launching the browser and thoroughly testing. By setting the safe search config option & safe search parameter to different levels, and by setting a regex in the blocklist and allowlist files. 

For example:

When safe search level is set to 3 or 4.

*In blocklist file*
```
book
```

*In allowlist file*
```
book[\t\n\r]+worm
```

## Author's checklist

- [x] add new config option `safe_search`.
- [x] add new URL search parameter `safesearch`. 
- [x] add new messages to show when results are filtered/disallowed.
- [x] add new images for the messages.
- [x] theme the messages provided. 
- [x] bump the app version to `v0.19.0` & crates version. 

## Related issues

Closes #201


alamin655 (dismissed):

alamin655 (approved):

PR #215, opened 2023-09-07
📝 Fix link to Documentation in Readme

Lissy93:
## What does this PR do?
No code changes, I just noticed that because the default branch is currently `rolling`, the link to the documentation was referencing `https://github.com/neon-mmd/docs/` which returns a 404. This update should fix that :)

(Have used `HEAD` instead of `rolling`, so that if you rename or change you default branch in the future, the link won't break)

Because it let's users read the docs without getting a 404 ;)

If you click the three dots (top left), then view file, then click the link to Documentation, it _should_ now resolve to the docs directory.

## Related issues

N/A 

P.S - your project looks awesome! Great work :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
   <b align="center"><a href="README.md">Readme</a></b> |
   <b><a href="https://discord.gg/SWnda7Mw5u">Discord</a></b> |
   <b><a href="https://github.com/neon-mmd/websurfx">GitHub</a></b> |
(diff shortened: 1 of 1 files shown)


neon-mmd (approved):
I believe even small changes can make a difference. So, thanks :heart: for working on this PR. We really appreciate it :+1: keep it up. I will merge this PR right away :rocket: :slightly_smiling_face: .

PR #223, opened 2023-09-10
✨  A four caching options for the search engine based on feature based compilation

gzsombor:
## What does this PR do?

This PR enables running the app without Redis (as a single, in-memory cache could be more than enough).
And moves the cache configuration into a central location, and the shared cache pushed down to the interested http routes.

I don't want to run yet another service (in case Redis), just to cache a couple of jsons for a minute.

Run the app – with commenting out the redis_url line in the config.lua. 

## Author's checklist

- [x] Added in memory cache implementation for the search engine.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ fake-useragent = {version="0.1.3"}
 env_logger = {version="0.10.0"}
 log = {version="0.4.20"}
 mlua = {version="0.8.10", features=["luajit"]}
-redis = {version="0.23.3", features=["tokio-comp","connection-manager"]}
+redis = {version="0.23.3", features=["tokio-comp","connection-manager"], optional = true}
 md5 = {version="0.7.0"}
 rand={version="0.8.5"}
 once_cell = {version="1.18.0"}
@@ -33,7 +33,7 @@ dhat = {version="0.3.2", optional = true}
 mimalloc = { version = "0.1.38", default-features = false }
 async-once-cell = {version="0.5.3"}
 actix-governor = {version="0.4.1"}
-mini-moka = "0.10"
+mini-moka = { version="0.10", optional = true}
 
 [dev-dependencies]
 rusty-hook = "^0.11.2"
@@ -67,4 +67,7 @@ rpath = false
 strip = "debuginfo"
 
 [features]
+default = ["in_memory_cache", "redis"]
 dhat-heap = ["dep:dhat"] 
+in_memory_cache = ["dep:mini-moka"]
+redis = ["dep:redis"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ dhat = {version="0.3.2", optional = true}
 mimalloc = { version = "0.1.38", default-features = false }
 async-once-cell = {version="0.5.3"}
 actix-governor = {version="0.4.1"}
+mini-moka = "0.10"
 
 [dev-dependencies]
 rusty-hook = "^0.11.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,7 +67,8 @@ rpath = false
 strip = "debuginfo"
 
 [features]
-default = ["in_memory_cache", "redis"]
+default = ["memory-cache"]
 dhat-heap = ["dep:dhat"] 
-in_memory_cache = ["dep:mini-moka"]
-redis = ["dep:redis"]
+memory-cache = ["dep:mini-moka"]
+redis-cache = ["dep:redis"]
+hybrid-cache = ["memory-cache", "redis-cache"]
--- a/websurfx/config.lua
+++ b/websurfx/config.lua
@@ -47,7 +47,6 @@ theme = "simple" -- the theme name which should be used for the website
 
 -- ### Caching ###
 redis_url = "redis://127.0.0.1:8082" -- redis connection url address on which the client should connect on.
-in_memory_cache = true
 
 -- ### Search Engines ###
 upstream_search_engines = {
--- a/websurfx/config.lua
+++ b/websurfx/config.lua
@@ -47,6 +47,7 @@ theme = "simple" -- the theme name which should be used for the website
 
 -- ### Caching ###
 redis_url = "redis://127.0.0.1:8082" -- redis connection url address on which the client should connect on.
+in_memory_cache = true
 
 -- ### Search Engines ###
 upstream_search_engines = {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -447,6 +447,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f30e7476521f6f8af1a1c4c0b8cc94f0bee37d91763d0ca2665f299b6cd8aec"
 
 [[package]]
+name = "bytecount"
+version = "0.6.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2c676a478f63e9fa2dd5368a42f28bba0d6c560b775f38583c8bbaa7fcd67c9c"
+
+[[package]]
 name = "byteorder"
 version = "1.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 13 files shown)


neon-mmd:
Thanks :heart: for working on such a wonderful PR, I really appreciate it Great work :+1:  but can you please elaborate the reason you want to include the `in-memory` cache implementation like `Redis` is a very fast caching server that's we have included it. So I would like why you might want to have an `in-memory` cache ? Like, is it faster than Redis ? Privacy reasons ? Or any other solid reason other than ease of usability ? :slightly_smiling_face:

gzsombor:
Thank you for your kind words!
My main reason is to reduce the unnecessary complexity of an extra service, both for coding - the requirement to convert data back and forth to json, waiting for network communications, etc. Or if  someone wants to write and run tests, they need to start an external Redis service, etc. And also it makes deployment more complex than absolutely needed. Especially, as currently, it is only used for a 60-sec long caching, without persistence, or anything more complicated, so for me, it doesn't make too much sense. If it were a Python or Ruby program, which is not great in memory management and in multi threading. 
I can understand, that an out-of-process, in memory cache could make sense, but in Rust, unless someone wants to run multiple websurfx servers against a common Redis, or really concerned, that they should be able to restart websurfx, without throwing out the cache, I can't think of a good reason, why I would use Redis. (And even in these things are required, probably they could just put a load-balancing webserver ahead of the websurfx instances, and it will be enough)
 So it's faster, easier to handle, and easier to deploy :smile:

neon-mmd:
> Thank you for your kind words! My main reason is to reduce the unnecessary complexity of an extra service, both for coding - the requirement to convert data back and forth to json, waiting for network communications, etc. Or if someone wants to write and run tests, they need to start an external Redis service, etc. And also it makes deployment more complex than absolutely needed. Especially, as currently, it is only used for a 60-sec long caching, without persistence, or anything more complicated, so for me, it doesn't make too much sense. If it were a Python or Ruby program, which is not great in memory management and in multi threading. I can understand, that an out-of-process, in memory cache could make sense, but in Rust, unless someone wants to run multiple websurfx servers against a common Redis, or really concerned, that they should be able to restart websurfx, without throwing out the cache, I can't think of a good reason, why I would use Redis. (And even in these things are required, probably they could just put a load-balancing webserver ahead of the websurfx instances, and it will be enough) So it's faster, easier to handle, and easier to deploy 😄

Thanks :heart:  for taking the time to explain everything. Though I still can not get the strong point from which I can get satisfied. Like, I loved the usability part and I totally agree with it, but I still don't see a clear point like `Redis` is also easy to set up. Just need to install it and launch it and done. Also, Redis is faster than in memory caching and performant and which aligns with our goal of the project. Moreover, in the future, we are planning to have it configurable by the user like the user can define how long they want to cache the results. So with that in mind, I still don't see a clear, strong reason to include this. Though your work is great, but we would like to know a strong point like something that can align with the goals of the project and I can instantly have it in the project :slightly_smiling_face: . I hope you got my point :slightly_smiling_face: .

gzsombor:
Honestly, I don't get, why this project needs Redis at all? What are the project goals for which Redis is required? Sorry, I haven't seen anything about this in the docs...  
And I'm absolutely sure, that Redis is not faster than in memory caching - it can't be. For Redis, we need to convert the structure to JSON, send the buffer over TCP and on the other end, Redis needs to put in a similar hashmap, just like what we would have with the in-process caching. And for retrieval, the same happens, but in the opposite direction

alamin655:
I agree with @gzsombor

neon-mmd:
Thanks :heart: for the explaination, I was about to leave a reply here actually about this. That I just did some research yesterday and I found this article (link below) which explains everything in great depth which I would also suggest you to read as well :slightly_smiling_face:.

https://www.lpalmieri.com/posts/caching-types-in-a-microservice-architecture/

After going through it I got your point :joy: like why you want to have this feature in the project. Also I discovered the  advantages of including it. So after knowing this I would be very glad having this feature into the project :slightly_smiling_face: .

One more thing before we get this merged, I would suggest having a feature based approach because you know if the user only wants to use `Redis` then we should not include the `Moka` crate and if the user uses `Memory Caching` then we don't want to include `Redis` crate because if we do include both then the compilation time will increase which is not  good as it creates bad user experience :slightly_smiling_face: .

Also, I would suggest 2 more features if possible, for `no cache` which will be useful for devices with low memory availability and another one `hybrid` which will be useful for ensuring high availability like for example if someone wants to use `Redis` provided via a VPS provider and for some reason it is down or not available then this will allow the search engine to automatically switch to using local caches until `Redis` server becomes available again. Also, I would suggest having this features done in separate PRs :slightly_smiling_face: . 

Once again thanks for making such wonderful PR :heart: we really appreciate it :+1: keep it up!!.

gzsombor:
Thank you for your understanding! :)
Adding feature flags wasn't too difficult, now, we have 'redis' and 'in_memory_cache'. One can build with both or without any of that, and during deployment, we can enable/disable as we like - or we can just disable any caching.

neon-mmd:
Thanks :heart: for the awesome work, but I would suggest having a few little changes, and then I will merge this PR right away :rocket: .

Cargo.toml
```diff
@@ -66,4 +67,7 @@ rpath = false
 strip = "debuginfo"
 
 [features]
+default = ["in_memory_cache", "redis"]
 dhat-heap = ["dep:dhat"] 
+in_memory_cache = ["dep:mini-moka"]
+redis = ["dep:redis"]
```
```suggestion
default = ["memory-cache"]
dhat-heap = ["dep:dhat"] 
memory-cache = ["dep:mini-moka"]
redis-cache = ["dep:redis"]
no-cache = []
hybrid-cache = ["memory-cache", "redis-cache"]
```
I would suggest having two more features, `hybrid-cache` and `no-cache`. Also, I would recommend only having the `memory-cache` to be included as a default feature because as you had discussed before, most people might not require the use of Redis at all and would be satisfied using the `memory-cache` feature. So I think only having the `memory-cache` would be better. What do you say?

websurfx/config.lua
```diff
@@ -47,6 +47,7 @@ theme = "simple" -- the theme name which should be used for the website
 
 -- ### Caching ###
 redis_url = "redis://127.0.0.1:8082" -- redis connection url address on which the client should connect on.
+in_memory_cache = true
```
```suggestion
```
I would suggest not having this option at all, I feel that having a feature based approach would be better :slightly_smiling_face: .

neon-mmd (changes requested):

Cargo.toml
gzsombor: The no-cache feature wont work, because these features are additive, so if you run:
```
cargo build  --features no-cache
```
That would still have the features specified in the default.
To have a build, without any caching, you need to build with:
```
cargo build  --no-default-features
```

Similarly, 
```
cargo build  --features redis-cache
```
will build with enabled redis-cache AND memory-cache too. 

I've renamed the features

websurfx/config.lua
gzsombor: That wont be too consistent.
In this setup, one can build the app, which supports:
* caching in Redis - when a `redis_url` is specified
* caching in process - when a `redis_url` is not specified, but the `in_memory_cache` set to true
* no caching - when there is no `redis_url` and `in_memory_cache` set to false

Cargo.toml
neon-mmd: Ok I see, thanks :heart: again for pointing this out. But I would like to make one last suggestion I don't know if it is good but I think I will still make it.

My suggestion is that we should take this approach `if it works` because I think this will solve the problem by covering all the four options.

For memory caching solution we can have it like this:

``` shell
cargo build 
```

This will as we know will enable the default feature `memory-cache`.

For hybrid caching solution we can have it like this:

``` shell
cargo build --feature redis-cache
```

For redis caching solution we can have it like this:

``` shell
cargo build --no-default-features --feature redis-cache
```

For no caching solution we can have it like this:

``` shell
cargo build --no-default-features
```

I think this approach will be better because this will eliminate the need to have the config option `in_memory_cache`. Also this will give us 3 advantages one it will reduce binary size which can be beneficial for any hosting on a low storage device, second remove unneccesary code like for example if we have memory caching solution enabled then this will make sure to remove the unneccesary redis cache code from the binary, finally third advantage is that build time will improve which is good you can see. That's why I am actually wanting to have feature based approach. 

Also, the reason why I am suggesting feature based approach is that you see the user will stick to only one caching solution when they will first build the app. It would be a very rare case when the user might move from one caching solution to other. 

Also, finally I think for the inconsistancy and complexity. I think we can explain all this build option in the `docs` which will make it easy for the user to follow and build the app. So with this in mind I don't think it would be a big issue. What do you say? :slightly_smiling_face:

neon-mmd:
Thanks :heart: for working on such a wonderful PR. We really appreciate it :+1: Since everything looks good, I will mege this PR right away :rocket: :slightly_smiling_face: .

neon-mmd (approved):
Thanks :heart: for working on such a wonderful PR. We really appreciate it :+1: Since everything looks good, I will merge this PR right away :rocket: :slightly_smiling_face: .

Also, I would also suggest/encourage you to update the `docs` in the next PR after I merge this PR so that everything remains in sync. I hope you will not mind working on it :slightly_smiling_face: .

PR #228, opened 2023-09-12
✨ A new theme `Arctic basalt` for the search engine

arctic-byte:
## What does this PR do?

This PR changes the UI & theme of `Websurfx`. 

> **Note**
> It breaks all the existing themes due to changes made in the HTML & CSS files.

This change is essential as it improves the UI of the search engine.

## Screenshots

image
image

## Author's checklist

- [x] Rewrite the HTML files.
- [x] Rewrite the simple theme.
- [x] Add new `arctic-basalt` colorscheme.


neon-mmd:
Thanks :heart: for working on such a wonderful PR. We really appreciate it :+1: keep it up!!

Also, the new theme looks amazing :rocket: and I love it, but I would like to have the following changes which I have listed down below so that it is easier to follow:
-  I would like to have the UI in such a way that I could have the same UI theme but with different colorschemes. Here is a demo with the old version of the `simple` theme:


https://github.com/neon-mmd/websurfx/assets/132049916/3d771c7b-f19c-45e8-bf90-bc0243b540eb

- I would like to have the dark/light toggle removed because if the theme's colorscheme can be changed, so I don't think there is any need for the dark/light toggle button. So I would suggest having it removed.
- Also, when I reviewed the files I have found unnecessary comments in the code, so I would recommend removing it because it unnecessarily increases the file size.
- I would also recommend having the fonts in one folder which makes the project more structured and organized like for example you can have the fonts grouped in a folder like `public/static/fonts/Vilonti/`.
- I would also recommend a re-addition of the other 12 colorschemes we had previously present in the project.
- Also, finally, I would recommend having the CSS variables present in the colorschemes files renamed back to the old style of naming because I think the new naming of the CSS variables will be too confusing for the normal user. Our projects' goal is to provide a high level of customizability, which means that we want to give the user the ability to easily create custom colorschemes on their own. 

I hope you will not mind making the following suggested set of changes :slightly_smiling_face: .

jcjms:
Hey, yeah sure. Thank you for your feedback. I agree with everything besides the naming of the css variables. For a normal user / dev it would 100% be easier to handle a descriptive css variable instead of something called 'color-x'. I mean, how would anyone know what the color is for? Even I as a UI dev got confused and didn't know where to implement them correctly. Therefore I would prefer not to make that change. But I will do follow-up commits with the changes you suggested in the next few days.

arctic-byte:
Hey, yeah sure. Thank you for your feedback. I agree with everything besides the naming of the css variables. For a normal user / dev it would 100% be easier to handle a descriptive css variable instead of something called 'color-x'. I mean, how would anyone know what the color is for? Even I as a UI dev got confused and didn't know where to implement them correctly. Therefore I would prefer not to make that change. But I will do follow-up commits with the changes you suggested in the next few days.

alamin655:
@neon-mmd I like the new colorscheme variable naming, so I would suggest not to change it back.

neon-mmd:
> @neon-mmd I like the new colorscheme variable naming, so I would suggest not to change it back.

Ok then, I think after reconsidering my decision I feel that the CSS naming looks good, and I think agree with alamin655. So I would suggest not changing the CSS variable names to the old naming scheme. :slightly_smiling_face:

PR #273, opened 2023-09-24
⚙️ Rewrite the search result parser code to make the engine code more concise

gzsombor:
## What does this PR do?

The current search engine implementation already contained a lot of repetitive code (setting up Selectors, locating the dom elements, etc), which can be refactored. This makes understanding one engine much more simple.

I want to easily add more search engines, and for that, it's important to not write too much code.

### DIFF
--- a/src/engines/searx.rs
+++ b/src/engines/searx.rs
@@ -14,11 +14,12 @@ use error_stack::{Report, Result, ResultExt};
 /// A new Searx engine type defined in-order to implement the `SearchEngine` trait which allows to
 /// reduce code duplication as well as allows to create vector of different search engines easily.
 pub struct Searx {
+    // The parser, used to interpret the search result.
     parser: SearchResultParser,
 }
 
 impl Searx {
-    // new Searchx engine
+    /// creates a Searx parser
     pub fn new() -> Result<Searx, EngineError> {
         Ok(Self {
             parser: SearchResultParser::new(
@@ -70,7 +71,7 @@ impl SearchEngine for Searx {
             &Searx::fetch_html_from_upstream(self, &url, header_map, request_timeout).await?,
         );
 
-        if let Some(no_result_msg) = document.select(&self.parser.no_result).nth(1) {
+        if let Some(no_result_msg) = self.parser.parse_for_no_results(&document).nth(1) {
             if no_result_msg.inner_html()
             == "we didn't find any results. Please use another query or search in more categories"
         {
@@ -79,33 +80,16 @@ impl SearchEngine for Searx {
         }
 
         // scrape all the results from the html
-        Ok(document
-            .select(&self.parser.results)
-            .map(|result| {
-                SearchResult::new(
-                    result
-                        .select(&self.parser.result_title)
-                        .next()
-                        .unwrap()
-                        .inner_html()
-                        .trim(),
-                    result
-                        .select(&self.parser.result_url)
-                        .next()
-                        .unwrap()
-                        .value()
-                        .attr("href")
-                        .unwrap(),
-                    result
-                        .select(&self.parser.result_desc)
-                        .next()
-                        .unwrap()
-                        .inner_html()
-                        .trim(),
-                    &["searx"],
-                )
+        self.parser
+            .parse_for_results(&document, |title, url, desc| {
+                url.value().attr("href").map(|url| {
+                    SearchResult::new(
+                        title.inner_html().trim(),
+                        url,
+                        desc.inner_html().trim(),
+                        &["searx"],
+                    )
+                })
             })
-            .map(|search_result| (search_result.url.clone(), search_result))
-            .collect())
     }
 }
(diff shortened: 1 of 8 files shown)


neon-mmd:
Thanks ❤️ for working on this gzsombor, I love the changes you made. Good job. Keep it up!! 👍🏻 We already were planning to have the code refactored for the search engines. So you are PR came at the right time, but before I merge this PR I would like to have a few minor changes. As well as, I would also like to know the reasoning behind one of the changes 🙂.

src/server/routes/search.rs
```diff
@@ -191,7 +191,7 @@ async fn results(
                     let engines: Vec<EngineHandler> = cookie_value
                         .engines
                         .iter()
-                        .filter_map(|name| EngineHandler::new(name))
+                        .filter_map(|name| EngineHandler::new(name).ok())
```
neon-mmd, line 194: Can you please provide the reason behind including this change ? Like I can see, one of the reasons that it helps standardize code which I really appreciate it, but when I tested the app on my local machine I found that it does not return any errors in the logs. So I feel that both the code of the previous and the new code work exactly the same. So I don't see any clear advantage between the two. That's why I would like to know like what was the reason behind having this change ?. 🙂

src/engines/searx.rs
```diff
@@ -3,16 +3,35 @@
...
 /// A new Searx engine type defined in-order to implement the `SearchEngine` trait which allows to
 /// reduce code duplication as well as allows to create vector of different search engines easily.
-pub struct Searx;
+pub struct Searx {
+    /// The parser, used to interpret the search result.
+    parser: SearchResultParser,
+}
+
+impl Searx {
+    /// creates a Searx parser
```
neon-mmd, line 22: The documentation looks good 👍🏻, but I would like to suggest elaborating it and making it in accordance to the `rust convention` which you can find it here:

 https://rust-lang.github.io/api-guidelines/documentation.html
 
 Also, I would suggest doing it for other functions too 🙂.
 
 Moreover, I would like to recommend taking a look at other files in the project as an example to write documentation following the rust convention. For example, I would suggest taking a look at this file as an example:
 
https://github.com/neon-mmd/websurfx/blob/rolling/src/cache/cacher.rs

neon-mmd (changes requested):

alamin655:
@Mergifyio update

src/server/routes/search.rs
gzsombor: Yes, sorry, it was my bad. 
The original idea was to fail early, during the engine initialization, if the selectors were incorrect, and I've just put that .ok() call to convert to an optional, but that's doesn't make too much sense.
So I've changed again, so if the engine initialization fails, the error will propagate upwards properly.

neon-mmd: Yes, thanks :heart: the new changes look much better and as such I am marking this as resolved :slightly_smiling_face: .

src/models/engine_models.rs
```diff
@@ -2,12 +2,14 @@
...
 use super::aggregation_models::SearchResult;
-use error_stack::{Result, ResultExt};
+use error_stack::{Report, Result, ResultExt};
 use std::{collections::HashMap, fmt, time::Duration};
 
 /// A custom error type used for handle engine associated errors.
 #[derive(Debug)]
 pub enum EngineError {
+    /// No matching engine found
+    EngineNotFound,
```
neon-mmd: I would suggest renaming this to `NoSuchEngineFound` because I think this would be both self-documenting. Also, it makes a lot more sense. What do you think? :slightly_smiling_face: .

src/models/engine_models.rs
```diff
@@ -24,6 +26,9 @@ pub enum EngineError {
 impl fmt::Display for EngineError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
+            EngineError::EngineNotFound => {
+                write!(f, "Search engine not found")
+            }
```
neon-mmd:
```suggestion
            EngineError::EngineNotFound(engine) => {
                write!(f, "No such engine with the name '{engine}' found")
            }
```

I would like to suggest another small change. Could you make the error message a little more verbose like including the engine name because this will help the user spot the issue more quickly and easily. :slightly_smiling_face:

neon-mmd (changes requested):

src/models/engine_models.rs
gzsombor: Yes, that's a good point!

src/models/engine_models.rs
gzsombor: I'm not good at naming things, so fine from me :)

src/engines/searx.rs
neon-mmd: Gzsombor, are you still working on this suggestion? If not then let us know 🙂 .

neon-mmd (approved):
Thanks ❤️ for working on such a cool PR. We really appreciate it 👍 . Since everything looks good. We will merge this PR as soon as possible. 🙂

alamin655 (approved):

PR #324, opened 2023-10-13
🔧 Build arguments to make writing GitHub action for auto releasing new docker images easier 

### LINKED ISSUE #321: 🔧 Production docker image with build args
### Description

On the current `Dockerfile` we cannot really specific if we want in memory or redis for the cache.
The idea would be to add build args to specific if we want to enable the feature or not.

### Screenshots

_No response_

Yes

### Additional information

_No response_

### LINKED ISSUE #322: ✨ Production docker image for arm64
### Description

Some people may want to host the service on a rasberryPI or similar, so it could be nice to have arm64 image ready to deploy!

### Screenshots

_No response_

Yes

### Additional information

_No response_

0xle0ne:
## What does this PR do?

This change the `Dockerfile` to enable different architecture and build args to select the cache type.
Some problem with cargo chef cache also got fix.
I also optimised the image to be smaller then before using musl as target and upx with a scratch image.
The final result is ~98% smaller then originally.

It will be easier to create a github workflow to release docker image for different architecture and cache type.

You can build the image using docker, this is the command i used:

```sh
docker build --build-arg CACHE=redis -t websurfx:1.0.14 .
```

## Author's checklist

I'm not quite sure what to add on this section, but i'll join a screenshot of the new image size.

image

## Related issues

Closes #322
Closes #321

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,40 +1,43 @@
-FROM rust:latest AS chef
+FROM --platform=$BUILDPLATFORM rust:1.73.0-alpine3.18 AS chef
 # We only pay the installation cost once,
 # it will be cached from the second build onwards
+RUN apk add --update alpine-sdk musl-dev g++ make libcrypto3 libressl-dev git upx perl build-base && rm -rf /var/cache/apk/*
 RUN cargo install cargo-chef --locked
 
 WORKDIR /app
 
 FROM chef AS planner
-COPY . . 
+COPY ./Cargo.toml ./Cargo.lock ./
 RUN cargo chef prepare --recipe-path recipe.json
 
-FROM chef AS builder
+FROM --platform=$BUILDPLATFORM chef AS builder
 COPY --from=planner /app/recipe.json recipe.json
-# Build dependencies - this is the caching Docker layer!
-# Uncomment the line below if you want to use the `hybrid` caching feature.
-# RUN cargo chef cook --release --features redis-cache --recipe-path recipe.json
-# Comment the line below if you don't want to use the `In-Memory` caching feature.
-RUN cargo chef cook --release --recipe-path recipe.json
-# Uncomment the line below if you want to use the `no cache` feature.
-# RUN cargo chef cook --release --no-default-features --recipe-path recipe.json
-# Uncomment the line below if you want to use the `redis` caching feature.
-# RUN cargo chef cook --release --no-default-features --features redis-cache --recipe-path recipe.json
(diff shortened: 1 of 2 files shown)


neon-mmd (changes requested):
Thanks :heart: for working on such a wonderful PR. We really appreciate it :+1: Keep it up!! But after testing the new docker image, I have discovered an issue that the local image does not seem to have an issuer certificate because of that when I make a search query it fails to get the results. Moreover, this issue didn't happen before in the previous versions of the docker images, So I would suggest investigating and fixing this issue. :slightly_smiling_face:

Also, here is a screenshot of the issue for more details:

image

0xle0ne:
The issue should be fixed

Dockerfile
```diff
@@ -1,40 +1,43 @@
-FROM rust:latest AS chef
+FROM --platform=$BUILDPLATFORM rust:1.73.0-alpine3.18 AS chef
 # We only pay the installation cost once,
 # it will be cached from the second build onwards
+RUN apk add --update alpine-sdk musl-dev g++ make libcrypto3 libressl-dev git upx perl build-base && rm -rf /var/cache/apk/*
```
neon-mmd:
```suggestion
RUN apk add --no-cache alpine-sdk musl-dev g++ make libcrypto3 libressl-dev git upx perl build-base
```

Also, I would like to suggest making a minor change like the new alpine image have recently add the `--no-cache` option which does some things automatically like removes the `/var/cache/apk/•` and updates the repository automatically. 🙂

neon-mmd (changes requested):

Dockerfile
0xle0ne: Ah yes i'm too used to the old way

neon-mmd (approved):
Thanks :heart: again for working on such a wonderful PR. Since everything looks good, I will merge this PR right away :rocket: :slightly_smiling_face: .

Also, once this PR is merged, I would suggest you to update the `installation docs` by adding a warning note to the user under these [section](https://github.com/neon-mmd/websurfx/blob/rolling/docs/installation.md#manual-deployment) at an appropriate place which should inform the user that we currently officially only support the x86-64 architectures. As such, we do not recommend you to deploy it on other architectures. But if you still want to do it, then do it at your own risk.

Also, here is a guide on how to add warning notes in Markdown, I would suggest following it:
https://github.com/orgs/community/discussions/16925

PR #335, opened 2023-10-15
✨ `Brave` for the search engine

### LINKED ISSUE #319: ✨ `Brave` for the search engine
Provide the `Brave` search engine for the `websurfx` search engine.  To start working on this issue, you will need to create a new file with the name of the desired engine in the folder `src/engines/` which is located under the codebase (`websurfx` directory). Then in the file, just create a new struct with the name of the engine and then create two `impl` blocks for the struct. One for implementing the `SearchEngine` trait on it and the other for implementing the `parsers` to be used for aggregating results.

> **Note**

JanKaczmarkiewicz:
## What does this PR do?

Implements `Brave` source of search results.

Allows users to search using `Brave` search engine.

Edit `config.lua` file to enable `Brave` source:
```lua
    ...
    Brave = true,
    ...
```
Run app: https://github.com/neon-mmd/websurfx#installation-and-testing-%EF%B8%8F

## Related issues

Closes #319

### DIFF
--- a/src/engines/brave.rs
+++ b/src/engines/brave.rs
@@ -62,7 +62,7 @@ impl SearchEngine for Brave {
             ("REFERER".to_string(), "https://google.com/".to_string()),
             (
                 "COOKIE".to_string(),
-                format!("safe_search={}", safe_search_level),
+                format!("safe_search={safe_search_level}"),
             ),
         ]))
         .change_context(EngineError::UnexpectedError)?;
--- a/src/engines/brave.rs
+++ b/src/engines/brave.rs
@@ -49,7 +49,7 @@ impl SearchEngine for Brave {
 
         let safe_search_level = match safe_search {
             0 => "off",
-            1 | 2 => "moderate",
+            1 => "moderate",
             _ => "strict",
         };
 
--- a/src/engines/brave.rs
+++ b/src/engines/brave.rs
@@ -1,0 +1,84 @@
+//! The `brave` module handles the scraping of results from the brave search engine
+//! by querying the upstream brave search engine with user provided query and with a page
+//! number if provided.
+
+use std::collections::HashMap;
+
+use reqwest::header::HeaderMap;
+use scraper::Html;
+
+use crate::models::aggregation_models::SearchResult;
+use error_stack::{Report, Result, ResultExt};
+
+use crate::models::engine_models::{EngineError, SearchEngine};
+
+use super::search_result_parser::SearchResultParser;
+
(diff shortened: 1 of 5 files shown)


src/engines/brave.rs
```diff
@@ -0,0 +1,84 @@
...
+    ) -> Result<HashMap<String, SearchResult>, EngineError> {
+        let url = format!("https://search.brave.com/search?q={query}&offset={page}");
+
+        let header_map = HeaderMap::try_from(&HashMap::from([
+            ("USER_AGENT".to_string(), user_agent.to_string()),
+            (
+                "CONTENT_TYPE".to_string(),
+                "application/x-www-form-urlencoded".to_string(),
+            ),
+        ]))
```
neon-mmd: Thanks :heart: for working on such a wonderful PR. We really appreciate it :+1: Keep it up!! But I would like to suggest a few changes, can you implement the safe-search functionality for this engine as well if possible. Also, can you add the `Referer` header as well, as by implementing this header it helps evade requests send by engine to be marked as a bot (in simple terms helps evade bot detection).

neon-mmd (changes requested):

src/engines/brave.rs
```diff
@@ -0,0 +1,95 @@
...
+        page: u32,
+        user_agent: &str,
+        request_timeout: u8,
+        safe_search: u8,
+    ) -> Result<HashMap<String, SearchResult>, EngineError> {
+        let url = format!("https://search.brave.com/search?q={query}&offset={page}");
+
+        let safe_search_level = match safe_search {
+            0 => "off",
+            1 | 2 => "moderate",
```
neon-mmd:
```suggestion
            1  => "moderate",
```

Ok I would suggest a final minor change, Actually the strict mode should be enabled for any `safe_search_level` value above 1. So I would recommend making this change and after this we will merge the PR. :slightly_smiling_face:

neon-mmd (changes requested):

src/engines/brave.rs
JanKaczmarkiewicz: lmkwyt

neon-mmd (commented):
Thanks :heart: again for working on this. Since everything looks good, I will merge this PR right away :rocket: :slightly_smiling_face: .

neon-mmd (approved):
Thanks :heart: again for working on this. Since everything looks good, I will merge this PR right away :rocket: :slightly_smiling_face: .

PR #336, opened 2023-10-16
✨ Docker compose for development

### LINKED ISSUE #320: ✨ Docker compose for development
### Description

Could be nice and smoother if we can start the project in development using `docker compose up`.
It need to have a specific `Dockerfile` with `rust` and `cargo watch` so everytime we update the code the build rerun inside the development container.

### Screenshots

_No response_

Yes

### Additional information

_No response_

0xle0ne:
## What does this PR do?

Added a development docker compose to quickly setup the project

This allow better reproductibility on different exploitation system (macos, windows)

You can test using this command:

```sh
docker compose up
```

## Related issues

Closes #320


README.md
```diff
@@ -141,6 +143,14 @@ redis-server --port 8082 &
...
 Once you have started the server, open your preferred web browser and navigate to <http://127.0.0.1:8080> to start using Websurfx.
 
+## Using docker
+
+If you are using docker you can quickly start the project for development purpose using docker compose.
+
+```sh
+docker compose up
+```
+
```
neon-mmd: I would like to suggest a minor change here, can you add this section in the [docs](https://github.com/neon-mmd/websurfx/tree/rolling/docs) with a new doc page dedicated to this and also explaining in depth about the process of setting docker deployment for development purposes. Moreover, if there is a need to add notes like `warning` or `notes` then please add them as well.

Here is an example of a dedicated `developing` doc page from the `dashy` project in their docs:

https://github.com/Lissy93/dashy/blob/master/docs/developing.md

websurfx/config.lua
```diff
@@ -5,7 +5,7 @@ threads = 10 -- the amount of threads that the app will use to run (the value sh
 
 -- ### Server ###
 port = "8080" -- port on which server should be launched
-binding_ip = "127.0.0.1" --ip address on the which server should be launched.
+binding_ip = "0.0.0.0" --ip address on the which server should be launched.
```
neon-mmd: Also, I would like to suggest not changing this file because this project is packaged in other packaging formats as well. When users will install these from these formats, they can mistakenly expose their IP address on the Internet because they might not know about it. Or they thought the default config was fine for running the search engine, which is, as you know, bad for privacy. So I would recommend not updating the file but providing a sample config with details on what to update and how to run the project for development in a separate page on the docs.

docker-compose.yml
```diff
@@ -1,18 +1,25 @@
----
```
neon-mmd: Also, I would like to suggest adding this back in the YAML file as this is considered a part of the YAML syntax. Moreover, I would also recommend updating the file from `docker-compose.yml` to `dev.docker-compose.yml` and for the `prod.docker-compose.yml` file I would recommend leaving it as it with the original `docker-compose.yml` name. I think this would be better. What do you say ? :slightly_smiling_face:

neon-mmd (changes requested):
Thanks :heart: for working on such a cool PR. We really appreciate it :+1: but I would like to suggest a few changes. :slightly_smiling_face:

README.md
0xle0ne: I feel like this is more your responsability to write the developing guide as you have more knowledge about the project then i have.

websurfx/config.lua
0xle0ne: Wdym by exposing IP address ?
Because it doesn't it just tell to the system to bind on all available IPV4 address on the local machine.
And there isn't other workarround to make it work on docker (production or development) also in my opinion theses kind of config should be ENV variables. Or passed as argument to the program directly.

docker-compose.yml
0xle0ne: Usually we have a docker-compose at the root of the project for development purpose.
And i'll remove the production one but that's up to you

neon-mmd: > Usually we have a docker-compose at the root of the project for development purpose. And i'll remove the production one but that's up to you

No actually, I am not suggesting you to remove the `prod.docker-compose.yml` file just renaming the files as I suggested above. Other than that, the development docker compose file and the development Dockerfile looks perfect :slightly_smiling_face: .

websurfx/config.lua
neon-mmd: > Wdym by exposing IP address ? Because it doesn't it just tell to the system to bind on all available IPV4 address on the local machine. And there isn't other workarround to make it work on docker (production or development) also in my opinion theses kind of config should be ENV variables. Or passed as argument to the program directly.

Yes, true, but then it also tells to bind on the IP address that was assigned to the system by the network. Let's say if it was a Public IP, then this could be an issue as you know this will accidentally expose the search engine publicly. :slightly_smiling_face: 

Also, I agree with you that env variables are a good idea, but they tend to be not very great when it comes to maintenance, flexibility, and customizability. For example, if you were going to host the search engine on four devices then you will need to put four different variables on each machine, which is, as you know, very painful. Moreover, if we were to let's say tomorrow propose a breaking change, then the user will have to set the new env variables on all four machines separately. So by using a config file and also one written in a programming language, it provides a good amount of flexibility. Since our config file is written in Lua, you can take advantage of loops, conditionals, etc. to write only one config which covers all the devices. Now if a change were to occur then this will require only the file needs to be changed. Also, there are benefits as well because of it being written in a programming language. I hope you got my point :slightly_smiling_face: .

0xle0ne: If you install it on a dedicated server, isn't the goal to have it publicly available?
You wont be able to access to it if you only bind to 127.0.0.1.
As for the ENV variable, it's industry standard to be able to set the HOST and the PORT.
And that most likely never causes a breaking change.
But whatever.

docker-compose.yml
0xle0ne: I understood what you asked, i'm just telling you what are the usual practice.

README.md
neon-mmd: I think you are right, no problem, we will update for these changes soon :slightly_smiling_face: .

neon-mmd (approved):
Thanks :heart: again for working on such a wonderful PR. We really appreciate it :+1: Since everything looks good, I will merge this PR right away :rocket: :slightly_smiling_face: .

PR #338, opened 2023-10-17
🤖 Improve the rules used by the `Mergify` bot.

alamin655:
## What does this PR do?

This PR improves the mergify bot configuration by updating the rules that need to be fulfilled before the bot merges the PR.

This change is essential as it improves the mergify bot workflow by checking for only certain critical checks to meet the conditions before the bot merges the PR automatically.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,7 +1,7 @@
 pull_request_rules:
   - name: Automatic merge on approval
     conditions:
-      - "#approved-reviews-by>=1"
+      - "#approved-reviews-by>=2"
       - check-success=triage
       - check-success=CodeFactor
       - check-success=Summary
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,8 +1,10 @@
 pull_request_rules:
   - name: Automatic merge on approval
     conditions:
-      - "status-success=checks/approved"
-      - "#approved-reviews-by>=2"
+      - "#approved-reviews-by>=1"
+      - check-success=triage
+      - check-success=CodeFactor
+      - check-success=Summary
     actions:
       queue:
         method: squash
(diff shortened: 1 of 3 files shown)


.mergify.yml
```diff
@@ -1,8 +1,10 @@
 pull_request_rules:
   - name: Automatic merge on approval
     conditions:
-      - "status-success=checks/approved"
-      - "#approved-reviews-by>=2"
+      - "#approved-reviews-by>=1"
+      - check-success=triage
+      - check-success=CodeFactor
+      - check-success=Summary
```
neon-mmd:
```suggestion
      - check-success=build (stable)
      - check-success=CodeFactor
      - check-success=Rust project
```

I would like to suggest having only the required checks to be set as a requirement to be fullfilled before the mergify bot merges the PR. Because I think the other checks like triage and summary even if they fail could still be handled by the maintainers manually. While other checks are important and can't be skipped, for example code quality check, build checks, etc. because they are important. Without them the project would either won't work or the code quality would not be great, which is as you know is never good. So with that in mind, I would like to recommend this minor change :slightly_smiling_face: .

.mergify.yml
```diff
@@ -1,8 +1,10 @@
 pull_request_rules:
   - name: Automatic merge on approval
     conditions:
-      - "status-success=checks/approved"
-      - "#approved-reviews-by>=2"
+      - "#approved-reviews-by>=1"
```
neon-mmd: I would suggest not changing this option and leaving it to 2 or more because I think having more reviews on a PR is always great because this gets more people's eyes on the changes that were made. Which can help get more ideas on the PR as wells as sometimes make the PR much better. :slightly_smiling_face:

neon-mmd (changes requested):
Thanks :heart: for working on this, but I would like to suggest a few changes before I merge this PR. :slightly_smiling_face:

neon-mmd (approved):
Thanks ❤️ for working on this. Since everything looks good, I will merge this PR right away 🚀 🙂 .

PR #396, opened 2023-11-26
:recycle: Clean some code, fix bug

ibnz36:
## What does this PR do?

Makes code cleaner, removes unnecessary double-function-calls, unnecessary mutability, and fixes an issue where the code would overflow in `src/server/routes/search.rs` if the `page` query parameter was 0.

It isn't extremely important, but the bugfix might come in handy.

All current tests work, and no big functionality was added, so using the current system works.

## Author's Additional Notes

Please open an issue to fix a massive performance concern: in the search server route, 3 pages worth of search results are being handled (the one requested, the one before and the one after) while only 1 is actually being sent to the user. Either send all 3, or make the other 2 pages load after the response is actually sent.

## Related issues

- None


neon-mmd (approved):
Thanks :heart: for working on such a wonderful PR. We really appreciate it :+1: Keep it up!! Since everything looks good. We will merge this PR right away 
:rocket:  :slightly_smiling_face:

PR #399, opened 2023-11-27
:recycle: Refactor cache system

### LINKED ISSUE #382: 🐛 Search results are no longer being cached in the `websurfx` search engine
### Description

After thorough testing, we have noticed that the search results are no longer being cached by both implementation of the caches the In-memory and Redis caching. After debugging the app, we receive two errors when the app tries to cache the results with both of them. The first, when the app tries to cache the result with Redis server, we receive an error stating something like `The response type is not of a valid string type`. Second, when the app tries to cache/retrieve results with the In-memory cache (mini-mocha), we get an error stating something like `The cache is missing the value`. 

### Screenshots

_No response_

No

### Additional information

_No response_

ashwinvin:
## What does this PR do?
This PR implements the following things:
- [x] Refactor the cache system to make it more testable and extensible.
- [x] Fix search results not being cached.

The current cache system is limited and is tightly coupled to just caching search results.
This PR abstracts the main logic and opens it up for newer cache backends which could be implemented.

`cargo test`

## Author's Additional Notes
Should admins be allowed to set the max number of search results to cache and TTL of search results?

## Related issues
Closes #382

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4066,7 +4066,7 @@ checksum = "1778a42e8b3b90bff8d0f5032bf22250792889a5cdc752aa0020c84abe3aaf10"
 
 [[package]]
 name = "websurfx"
-version = "1.2.30"
+version = "1.2.34"
 dependencies = [
  "actix-cors",
  "actix-files",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "websurfx"
-version = "1.2.30"
+version = "1.2.34"
 edition = "2021"
 description = "An open-source alternative to Searx that provides clean, ad-free, and organic results with incredible speed while keeping privacy and security in mind."
 repository = "https://github.com/neon-mmd/websurfx"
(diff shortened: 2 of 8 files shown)


neon-mmd:
The changes so far looks great :+1:  I think the trait based approach would be better because I see your point having this approach would be more extensible. So this a completely `yes` from my side. :slightly_smiling_face: 

---

> Should admins be allowed to set the max number of search results to cache and TTL of search results?

Also, about how many results could be stored in In-Memory caching solution and the expiry of results in both the caching solutions. I think it would be better to provide control for this to the admin as the admin would have control on how much memory to allocate as well as for how long cache should remain valid. For these options we could provide to options in the config and for this work I think we can have a separate issue for this. :slightly_smiling_face:

ashwinvin:
Its pretty easy to add the configuration settings for that. Yeah, lets put up a new issue to add those settings once this is merged.

neon-mmd (approved):
Thanks :heart: for working on such a wonderful PR. We really appreciate it :+1: Keep it up!! Since everything looks good. We will merge this PR right away :rocket: :slightly_smiling_face: .

PR #403, opened 2023-11-30
✨ Config option to customize the cache invalidation/expiry time

### LINKED ISSUE #402: ✨ Config option to customize the cache invalidation/expiry time
### Description

Provide a new config option to customize the cache invalidation or expiry of the cached result in the `config.lua` file located in the `websurfx` directory located under the codebase (`websurfx` directory). 

The new config option might look something like this:
``` lua
cache = {
	expiry_time = 60 -- This option takes the expiry time of the search results (value in seconds and the value should be greater than or equal to 60 seconds).
}
```

> [!Note]
> 1. When implementing the parsing of the new config option, make sure to implement proper checks to eliminate unexpected behaviors, like implementing checks to eliminate the case where user by mistakenly sets a value less than 60 in the config file. 
> 2. Only allow parse this option when any of the three caching features other than `no-cache` caching feature is enabled. 

### Screenshots

_No response_

None

### Additional information

_No response_

ashwinvin:
## What does this PR do?

Provides a new config option `cache_expiry_time`. 

Allows the admin to set TTL (time to live) value for cached search results. 

## Related issues

Closes #402

### DIFF
--- a/src/cache/redis_cacher.rs
+++ b/src/cache/redis_cacher.rs
@@ -18,6 +18,8 @@ pub struct RedisCache {
     pool_size: u8,
     /// It stores the index of which connection is being used at the moment.
     current_connection: u8,
+    /// It stores the max TTL for keys.
+    cache_ttl: u16,
 }
 
 impl RedisCache {
@@ -36,6 +38,7 @@ impl RedisCache {
     pub async fn new(
         redis_connection_url: &str,
         pool_size: u8,
+        cache_ttl: u16,
     ) -> Result<Self, Box<dyn std::error::Error>> {
         let client = Client::open(redis_connection_url)?;
         let mut tasks: Vec<_> = Vec::new();
@@ -48,6 +51,7 @@ impl RedisCache {
             connection_pool: try_join_all(tasks).await?,
             pool_size,
             current_connection: Default::default(),
+            cache_ttl,
         };
         Ok(redis_cache)
     }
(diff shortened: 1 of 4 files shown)


src/cache/redis_cacher.rs
```diff
@@ -18,6 +18,8 @@ pub struct RedisCache {
     pool_size: u8,
     /// It stores the index of which connection is being used at the moment.
     current_connection: u8,
+    /// It stores the max TTL for keys.
+    cache_ttl: u16,
```
neon-mmd, line 22: I think I would suggest not having the TTL value stored in the struct because we are going to use the TTL value only once when we are setting the cache in the Redis server. So by having the TTL value stored in the struct, it consumes unnecessary resources of the system. :slightly_smiling_face:

neon-mmd (changes requested):
Thanks :heart: for working on this. We really appreciate it :+1: but before we get this PR merged, I would suggest a few changes.

src/cache/redis_cacher.rs
ashwinvin: It is just two bytes which is really small and the same argument could be said for `threads` since it is only used once while initializing the actix server.

neon-mmd: > It is just two bytes which is really small and the same argument could be said for `threads` since it is only used once while initializing the actix server.

I think you are right let's have this then :slightly_smiling_face: .

neon-mmd (approved):
Thanks :heart: again for working on this. Since everything looks good. I will merge this PR right away :rocket: :slightly_smiling_face: .

PR #415, opened 2023-12-02
🎨 Restructure search engine backend

ashwinvin:
## What does this PR do?

This PR does the following:
- Repurposed the`EngineHandler` struct to schedule and handle engine execution and related processes
- Introduced `Ranker` struct to handle result aggregation and ranking.
- Introduced the `HttpClient` struct meant to handle all the http requests made and added configuration for it.
- Introduce new methods for `SearchEngine` trait to get metadata about engine. (`get_query_types`, `get_name`)
- Added enums to support more options to go with query (Relavancy and Type) (Not implemented in functions as of now).
- Made error handling a bit more concise.

This PR is decouples the components of backend and also allow for future extension of the objects.

`cargo run`

## Author's checklist

- [ ] Add documentation (need help)
- [ ] Complete the TODO's 
- [ ] Implement checks for proxy & tor connection

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ bench = false
 path = "src/bin/websurfx.rs"
 
 [dependencies]
-reqwest = {version="0.11.22", default-features=false, features=["rustls-tls","brotli", "gzip"]}
+reqwest = {version="0.11.22", default-features=false, features=["rustls-tls","brotli", "gzip", "json"]}
 tokio = {version="1.32.0",features=["rt-multi-thread","macros"], default-features = false}
 serde = {version="1.0.190", default-features=false, features=["derive"]}
 serde_json = {version="1.0.108", default-features=false}
@@ -38,6 +38,8 @@ mimalloc = { version = "0.1.38", default-features = false }
 async-once-cell = {version="0.5.3", default-features=false}
 actix-governor = {version="0.5.0", default-features=false}
 mini-moka = { version="0.10", optional = true, default-features=false, features=["sync"]}
+enumflags2 = "0.7.8"
+time ={version = "0.3.30", features = ["serde-human-readable"]}
 
 [dev-dependencies]
 rusty-hook = {version="^0.11.2", default-features=false}
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1,0 +1,91 @@
+use std::{collections::HashMap, sync::Arc};
+
+use crate::models::{
+    aggregation_models::SearchResult,
+    client_models::HttpClient,
+    engine_models::{EngineError, EngineErrorType, SearchEngine},
+};
+use actix_web::rt::spawn;
+use error_stack::Report;
+use error_stack::Result;
+
+pub struct EngineHandler {
+    engines: Arc<Vec<Arc<Box<dyn SearchEngine>>>>,
+    client: Arc<HttpClient>,
+}
+
+pub type RawResults = Vec<Result<HashMap<String, SearchResult>, EngineError>>;
+
+impl EngineHandler {
+    /// Parses names of engines and initialises them for use.
+    ///
+    /// # Arguments
+    ///
+    /// * `engine_names - It takes the names of the engines.
+    ///
+    /// # Returns
+    ///
+    /// It returns an option either containing the value or a none if the engine is unknown
+    pub fn new(engine_names: Vec<String>, client: HttpClient) -> Result<Self, EngineError> {
+        let mut engines: Vec<Arc<Box<dyn SearchEngine>>> = vec![];
+
+        for engine_name in engine_names {
+            let engine: Box<dyn SearchEngine> = match engine_name.to_lowercase().as_str() {
+                "duckduckgo" => Box::new(crate::engines::duckduckgo::DuckDuckGo::new()?),
+                "searx" => Box::new(crate::engines::searx::Searx::new()?),
+                "brave" => Box::new(crate::engines::brave::Brave::new()?),
+                _ => {
+                    return Err(Report::from(EngineError {
+                        error_type: EngineErrorType::NoSuchEngineFound,
+                        engine: engine_name.to_string(),
+                    }))
+                }
+            };
+            engines.push(Arc::new(engine));
+        }
+
+        Ok(Self {
+            engines: Arc::new(engines),
+            client: Arc::new(client)
+        })
+    }
+
+    pub async fn search(
+        &self,
+        engine_names: Option<Vec<String>>,
+        query: &str,
+        // category: QueryType,
+        // query_relevance: Option<QueryRelavancy>,
+        page: u32,
+        safe_search: u8,
+    ) -> RawResults {
+        let mut tasks = Vec::with_capacity(self.engines.len());
+        for engine in &*self.engines {
+            if let Some(ref engine_names) = engine_names {
+                // TODO: Handle invalid engine names provided by the user, currently it silently ignores
+                if !engine_names.contains(&engine.get_name().to_owned()) {
+                    continue;
+                }
+            }
+            let engine = engine.clone();
+            // let query_relevance = query_relevance.clone();
+            let client = self.client.clone();
+            let query = query.to_owned();
+            tasks.push(spawn(async move {
+                engine
+                    .fetch_results(&query, page, client, safe_search)
+                    .await
+            }));
+        }
+
+        let mut responses = Vec::with_capacity(tasks.len());
+        for task in tasks {
+            // An error will only be raised when the task panics, here it should technically never panic
+            if let Ok(result) = task.await {
+                responses.push(result);
+            }
+        }
+
+        responses
+    }
+}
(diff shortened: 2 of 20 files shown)


Cargo.toml
```diff
@@ -38,6 +38,8 @@ mimalloc = { version = "0.1.38", default-features = false }
 async-once-cell = {version="0.5.3", default-features=false}
 actix-governor = {version="0.5.0", default-features=false}
 mini-moka = { version="0.10", optional = true, default-features=false, features=["sync"]}
+enumflags2 = "0.7.8"
+time ={version = "0.3.30", features = ["serde-human-readable"]}
```
neon-mmd, line 42: I would suggest not including this crate and instead using the Unix epoch method to compute random delay between requests

neon-mmd, line 42:
```suggestion
```

I would suggest not including this crate and instead using the Unix epoch method to compute random delay between requests. By not including the crate, it helps reduce the amount of crates that is required to compile the app, which is essential for a good user experience. :slightly_smiling_face:

websurfx/config.lua
```diff
@@ -15,6 +15,16 @@ rate_limiter = {
...
+request_client = {
+	-- proxy_url =""
+	-- is_tor_proxy = "false"
+	use_http2 = "true",
+	https_only = "true",
+	timeout = "5",
+	max_retries = "3",
+	max_redirects = "3"
+}
+
```
neon-mmd, line 27:
```suggestion
```

I would suggest not including this in the config because this is something we don't want to include as of now :slightly_smiling_face:

src/models/aggregation_models.rs
```diff
@@ -15,11 +18,42 @@ pub struct SearchResult {
...
+    },
+    FileSearchResult {
+        /// The metadata regarding the file. eg: author, size, etc
+        metadata: HashMap<String, String>,
+        /// The time of upload of the file.
+        uploaded_at: Option<Date>,
+    },
+    AutoCompletionResult {
+        /// The suggested completion by the upstream search engine.
+        suggestion: String,
```
neon-mmd, line 55:
```suggestion
```

I would suggest not including this code because this is a feature we plan to include it in the future. So again we don't want to include it as of now 🙂

src/models/client_models.rs
```diff
@@ -0,0 +1,120 @@
...
+
+#[derive(Debug)]
+pub enum ClientError {
+    /// Raised when a web request is failed.
+    RequestError,
+    /// Raised when the proxy url is invalid.
+    InvalidProxy,
+    /// Raised when the proxy is marked as tor enabled but is not routing the requests
+    /// through tor exit nodes.
+    TorInactive,
```
neon-mmd, line 17:
```suggestion
```

I would suggest not including this code because this is a feature we plan to include it in the future. So again we don't want to include it as of now 🙂

src/models/engine_models.rs
```diff
@@ -26,65 +33,56 @@ pub enum EngineError {
...
+}
+
+#[derive(Debug, Clone)]
+pub enum QueryRelavancy {
+    Anytime,
+    LastDay,
+    LastWeek,
+    LastMonth,
+    LastYear,
+}
```
neon-mmd, line 78:
```suggestion
```

I would suggest not including this code because this is a feature we plan to include it in the future. So again we don't want to include it as of now 🙂

src/models/parser_models.rs
```diff
@@ -48,3 +50,56 @@ pub struct RateLimiter {
...
+        Ok(RequestClientConfig {
+            proxy_url: value.get("proxy_url").cloned(),
+            is_tor_proxy,
+            https_only,
+            use_http2: value["use_http2"].parse()?,
+            timeout: value["timeout"].parse()?,
+            max_retries: value["max_retries"].parse()?,
+            max_redirects: value["max_redirects"].parse()?,
+        })
+    }
```
neon-mmd, line 104:
```suggestion
```

I would suggest not including this code because this is a feature we plan to include it in the future. So again, we don't want to include it as of now 🙂. Also, one more thing to note as the project has a strong goal towards security, therefore we would strongly recommend only having the `https_only` mode. As this will provide the user with results over a secure HTTPS connection.

src/results/aggregator.rs
```diff
@@ -1,186 +1,111 @@
...
-            engine_name,
-        ));
-    };
-
-    for _ in 0..responses.len() {
-        let response = responses.pop().unwrap();
-        let engine = names.pop().unwrap();
+pub struct Ranker;
+
+impl Ranker {
```
neon-mmd, line 23:
```suggestion
pub struct Aggregator;

impl Aggregrator {
```

I would suggest renaming this struct to Aggregator, as we feel having the name Ranker would be very confusing for anybody working on the project. As there is currently a work for having a ranking algorithm being worked on which can cause confusion between the two. I hope you got my point :slightly_smiling_face:

src/engine.rs
```diff
@@ -0,0 +1,91 @@
+use std::{collections::HashMap, sync::Arc};
+
+use crate::models::{
+    aggregation_models::SearchResult,
+    client_models::HttpClient,
+    engine_models::{EngineError, EngineErrorType, SearchEngine},
+};
+use actix_web::rt::spawn;
```
neon-mmd, line 8: I would suggest not using the `actix-web's` spawn function as it does not provide the same functionality as `tokio's` spawn function. `Tokio` spawn creates both concurrent and parallel, which can be very advantages for having really quick i/o bound tasks to be performed.

For more information on this, I would suggest watching this video, which explains in depth how `tokio` spawn works:

https://piped.video/ThjvMReOXYM?feature=shared&t=4767

src/lib.rs
```diff
@@ -7,6 +7,7 @@
 
 pub mod cache;
 pub mod config;
+pub mod engine;
 pub mod engines;
```
neon-mmd, line 11: I would suggest having a different name for the engine file, as it would be very confusing for anyone working on the project to figure out what is the difference between the two. :slightly_smiling_face:

src/lib.rs
```diff
@@ -76,6 +83,8 @@ pub fn run(
             .wrap(Logger::default()) // added logging middleware for logging.
             .app_data(web::Data::new(config.clone()))
             .app_data(cache.clone())
+            .app_data(engine_handler.clone())
+            .app_data(ranker.clone())
```
neon-mmd, line 87: I think I would suggest not having the two structs initialized with app data as it would consume more resources and is not ideal and not very efficient solution :slightly_smiling_face: .

src/models/client_models.rs
```diff
@@ -0,0 +1,120 @@
...
+        header_map: reqwest::header::HeaderMap,
+        timeout: Option<u8>,
+    ) -> Result<String, ClientError> {
+        self.fetch(url, header_map, timeout)
+            .await?
+            .json()
+            .await
+            .change_context(ClientError::SerializationError)
+    }
+}
```
neon-mmd, line 120:
```suggestion
```

I would strongly suggest not including this, as we will not and don't want to support fetching results as JSON. As this can make maintaining the project more difficult as there will be a more way to fetch data and would become very complex and not very standardized. :slightly_smiling_face:

src/models/client_models.rs
```diff
@@ -0,0 +1,120 @@
...
+        url: &str,
+        header_map: reqwest::header::HeaderMap,
+        timeout: Option<u8>,
+    ) -> Result<String, ClientError> {
+        self.fetch(url, header_map, timeout)
+            .await?
+            .text()
+            .await
+            .change_context(ClientError::RequestError)
+    }
```
neon-mmd, line 106:
```suggestion
    pub async fn fetch_html(
        &self,
        url: &str,
        header_map: reqwest::header::HeaderMap,
        timeout: Option<u8>,
    ) -> Result<String, ClientError> {
        self.fetch(url, header_map, timeout)
            .await?
            .text()
            .await
            .change_context(ClientError::RequestError)
         tokio::sleep(random_delay).await?
    }
```

I would strongly suggest having a random delay after a request as we don't want to accidentally DDoS the upstream search engine :slightly_smiling_face:

neon-mmd:
Thanks ❤️ for working on this PR, Great work :+1:  but since this PR includes many changes that are planned to be included in the future. Also, since it includes many changes which completely changes the search engine's backend like removing random delay which can negatively impact the search engine by DDoSing the upstream servers. Moreover, removing user_agent header from the requests which can negatively impact the server by getting it blocked due to bot detection, etc. as discussed above. Therefore, Sorry, but we would not accept this PR. 🙂. But we hope you will not mind. Hope to continue seeing you with cool PRs like the previous one in the future. 🙂

neon-mmd:
Sorry for the misunderstanding, we thought this is a normal PR and closed it but we will make sure this does not happen again. Also, Thanks :heart: again for working on such a wonderful PR. Hope to continue seeing the cool work :rocket: :slightly_smiling_face: .

neon-mmd (changes requested):

PR #419, opened 2023-12-05
✨ `Startpage` for the search engine

### LINKED ISSUE #314: ✨ `Startpage` for the search engine
Provide the `startpage` search engine for the `websurfx` search engine.  To start working on this issue, you will need to create a new file with the name of the desired engine in the folder `src/engines/` which is located under the codebase (`websurfx` directory). Then in the file, just create a new struct with the name of the engine and then create two `impl` blocks for the struct. One for implementing the `SearchEngine` trait on it and the other for implementing the `parsers` to be used for aggregating results.

> **Note**

neon-mmd:
## What does this PR do?

This PR provides the `startpage` upstream search engine.

This PR is essential, as adding more engines will provide more diverse search results for the user, which improves the user experience.

It can be tested by installing and running `Websurfx` as mentioned in the `docs` and on the `readme` and by launching the browser and thoroughly testing. By selecting the Startpage search engine from the settings page or by setting it in the config. Then checking whether the results are being provided from the Startpage search engine or not.

## Author's checklist

- [x] Provide the Startpage upstream search engine
- [x] Provide the Startpage engine as an option in the config 
- [x] Bump the app version to `v1.4.0` 

## Related issues

Closes #314


alamin655 (approved):

PR #429, opened 2023-12-08
✨ `LibreX` for the search engine

### LINKED ISSUE #318: ✨ `LibreX` for the search engine
Provide the `LibreX` search engine for the `websurfx` search engine.  To start working on this issue, you will need to create a new file with the name of the desired engine in the folder `src/engines/` which is located under the codebase (`websurfx` directory). Then in the file, just create a new struct with the name of the engine and then create two `impl` blocks for the struct. One for implementing the `SearchEngine` trait on it and the other for implementing the `parsers` to be used for aggregating results.

> **Note**

alamin655:
## What does this PR do?

This PR provides the `librex` upstream search engine.

This PR is essential, as adding more engines will provide more diverse search results for the user, which improves the user experience.

It can be tested by installing and running `Websurfx` as mentioned in the `docs` and on the `readme` and by launching the browser and thoroughly testing. By selecting the LibreX search engine from the settings page or by setting it in the config. Then checking whether the results are being provided from the LibreX search engine or not.

## Author's checklist

- [x] Provide the LibreX upstream search engine
- [x] Provide the LibreX engine as an option in the config 
- [x] Bump the app version to `v1.5.0` 

## Related issues

Closes #318

### DIFF
--- a/src/engines/librex.rs
+++ b/src/engines/librex.rs
@@ -33,9 +33,9 @@ impl LibreX {
             parser: SearchResultParser::new(
                 ".text-result-container>p",
                 ".text-result-container",
-                ".text-result-container>a>h2",
-                ".text-result-container>a",
-                ".text-result-container>span",
+                ".text-result-wrapper>a>h2",
+                ".text-result-wrapper>a",
+                ".text-result-wrapper>span",
             )?,
         })
     }
--- a/src/engines/librex.rs
+++ b/src/engines/librex.rs
@@ -1,4 +1,4 @@
-/// The `LibreX` module contains the implementation of a search engine for LibreX using the reqwest and scraper libraries.
+/// The `librex` module contains the implementation of a search engine for LibreX using the reqwest and scraper libraries.
 /// It includes a `SearchEngine` trait implementation for interacting with the search engine and retrieving search results.
 
 use std::collections::HashMap;
--- a/src/engines/librex.rs
+++ b/src/engines/librex.rs
@@ -1,0 +1,102 @@
+/// # LibreX Search Engine
+///
+/// The `LibreX` module contains the implementation of a search engine for LibreX using the reqwest and scraper libraries.
+/// It includes a `SearchEngine` trait implementation for interacting with the search engine and retrieving search results.
+
+use std::collections::HashMap;
+
+use reqwest::header::HeaderMap;
+use reqwest::Client;
+use scraper::Html;
+
+use crate::models::aggregation_models::SearchResult;
+use crate::models::engine_models::{EngineError, SearchEngine};
+
+use error_stack::{Report, Result, ResultExt};
+
+use super::search_result_parser::SearchResultParser;
+
+/// Represents the LibreX search engine.
+pub struct LibreX {
+    /// The parser used to extract search results from HTML documents.
+    parser: SearchResultParser,
+}
+
+impl LibreX {
+    /// Creates a new instance of LibreX with a default configuration.
+    ///
+    /// # Returns
+    ///
+    /// Returns a `Result` containing `LibreX` if successful, otherwise an `EngineError`.
+    pub fn new() -> Result<Self, EngineError> {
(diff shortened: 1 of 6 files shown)


src/engines/librex.rs
```diff
@@ -0,0 +1,102 @@
...
+            _ => format!("https://search.ahwx.org/search.php?q={query}&p={}&t=10", page * 10),
+        };
+        
+        let header_map = HeaderMap::try_from(&HashMap::from([
+            ("USER_AGENT".to_string(), user_agent.to_string()),
+            ("REFERER".to_string(), "https://google.com/".to_string()),
+            ("CONTENT_TYPE".to_string(), "application/x-www-form-urlencoded".to_string()),
+            (
+                "COOKIE".to_string(),
+                "theme=dark; language=en; number_of_results=10; safe_search=on; save=1".to_string(),
```
neon-mmd:
```suggestion
                "theme=amoled; disable_special=on; disable_frontends=on; language=en; number_of_results=10; safe_search=on; save=1".to_string(),
```

I would suggest using this cookie instead, as this seems to provide better results 🙂.

src/engines/librex.rs
```diff
@@ -0,0 +1,102 @@
...
+            parser: SearchResultParser::new(
+                ".text-result-container>p",
+                ".text-result-container",
+                ".text-result-container>a>h2",
+                ".text-result-container>a",
+                ".text-result-container>span",
+            )?,
+        })
+    }
+}
```
neon-mmd:
```suggestion
    pub fn new() -> Result<Self, EngineError> {
        Ok(Self {
            parser: SearchResultParser::new(
                ".text-result-container>p",
                ".text-result-container",
                ".text-result-wrapper>a>h2",
                ".text-result-wrapper>a",
                ".text-result-wrapper>span",
            )?,
        })
    }
}
```

I would suggest using these selectors instead, as the previous ones on testing didn't work as intended and didn't provide the correct results. :slightly_smiling_face:

src/engines/librex.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// * `query` - The search query.
+    /// * `page` - The page number for pagination.
+    /// * `user_agent` - The user agent string.
+    /// * `client` - The reqwest client for making HTTP requests.
+    /// * `_safe_search` - A parameter for safe search (not currently used).
+    ///
+    /// # Returns
+    ///
+    /// Returns a `Result` containing a `HashMap` of search results if successful, otherwise an `EngineError`.
+    #[allow(clippy::unnecessary_wraps)] // The `Err` variant is explicit for better documentation.
```
neon-mmd:
```suggestion
   // The `Err` variant is explicit for better documentation.
```

I would suggest removing this Clippy check from the code as it is no longer required. :slightly_smiling_face:

neon-mmd (changes requested):
Thanks ❤️ for working on this. We really appreciate it 👍 but before we merge this PR, I would like to suggest a few changes. 🙂

src/engines/librex.rs
```diff
@@ -0,0 +1,111 @@
+/// The `librex` module contains the implementation of a search engine for LibreX using the reqwest and scraper libraries.
+/// It includes a `SearchEngine` trait implementation for interacting with the search engine and retrieving search results.
```
neon-mmd:
```suggestion
//! The `librex` module contains the implementation of a search engine for LibreX using the reqwest and scraper libraries.
//! It includes a `SearchEngine` trait implementation for interacting with the search engine and retrieving search results.
```

One final suggestion, before we merge this PR in, I would suggest replacing the third slash in the comment with an exclamation because module comments in rust uses `\\!` syntax. 🙂

neon-mmd (changes requested):

neon-mmd (approved):
Thanks ❤️ again for working on this. Since everything looks good I will merge this PR right away 🚀 🙂.

PR #443, opened 2023-12-16
✨ Compression and encryption for the cached search results

### LINKED ISSUE #416: ✨ Compression and encryption for the cached search results
#### Proposal of the Changes

Provide compression, encryption and both for the search results that are going to be a part of the cache by using the rust's feature/conditional based compilation feature. 

The three features should look something this: 
- `Default`: No compression and encryption
- `Compress-cache-results:`  Uses compression algorithm like `brotli` (which is considered good for text based compression)
- `Encrypt-cache-results:` Uses an encryption algorithm like `chacha20poly1305` (which is considered very strong encryption and is widely adopted in the `rust` programming language). Also, the generated encryption should change on each restart of the `websurfx` server, which help against attacks and improve the privacy of the user. 
-  `cec-cache-results:` Uses both compression and encryption to first compress the results then encrypt it and then compress it back again to get the advantages of both encryption and compression. 

### Screenshots

_No response_

None

#### The Reason Behind the Proposal

The reasoning behind the changes are as follows : 
- By compressing cache results, this can help reduce the amount of memory used for storing the cache results. Second, it can help improve the speed of transferring cache results over the internet. The size of the cache results become smaller after compression. So, if the cache results were to be stored on a Redis server provided by a VPS provider.
- By encrypting cache results, this can help improve user privacy by not allowing the Redis VPS provider to know what is being cached on their Redis server. Though, the cache results size become larger after encryption.
- By using both the encryption and compression with `cec-cache-results` feature, we provide the advantages of both compression and encryption by first compressing the results and the encrypting it. Then compressing it again so that the size of the cache result remains nearly the same as it was after the first compression.

#### Sample Code

Here is the sample which provides a general idea of what should be used to implement the above discussed features:

``` rust
use base64::Engine;
use chacha20::{
    cipher::{
        generic_array::GenericArray,
        typenum::{UInt, UTerm},
        StreamCipherCoreWrapper,
    },
    ChaChaCore,
};
use chacha20poly1305::{
    aead::Aead,
    aead::OsRng,
    consts::{B0, B1},
    AeadCore, ChaCha20Poly1305, ChaChaPoly1305, KeyInit,
};
use std::{io::Write, sync::OnceLock};

static CIPHER: OnceLock<
    ChaChaPoly1305<
        StreamCipherCoreWrapper<ChaChaCore<UInt<UInt<UInt<UInt<UTerm, B1>, B0>, B1>, B0>>>,
    >,
> = OnceLock::new();

static ENCRYPTION_KEY: OnceLock<GenericArray<u8, UInt<UInt<UInt<UInt<UTerm, B1>, B1>, B0>, B0>>> =
    OnceLock::new();

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let some_content = r#"
{
	"random": "20",
	"random float": "19.309",
	"bool": "true",
	"date": "1996-10-14",
	"regEx": "hellooooo to you",
	"enum": "online",
	"firstname": "Dorthy",
	"lastname": "Dreda",
	"city": "Forked River",
	"country": "Germany",
	"countryCode": "ZW",
	"email uses current data": "Dorthy.Dreda@gmail.com",
	"email from expression": "Dorthy.Dreda@yopmail.com",
	"array": [
		"Quintina",
		"Courtnay",
		"Ida",
		"Chloris",
		"Regina"
	],
	"array of objects": [
		{
			"index": "0",
			"index start at 5": "5"
		},
		{
			"index": "1",
			"index start at 5": "6"
		},
		{
			"index": "2",
			"index start at 5": "7"
		}
	],
	"Elka": {
		"age": "39"
	}
}
    "#;

    let mut writer = brotli::CompressorWriter::new(Vec::new(), 4096, 11, 22);
    writer.write_all(some_content.as_bytes())?;
    let something = writer.into_inner();

    let cipher = CIPHER.get_or_init(|| {
        let key = ChaCha20Poly1305::generate_key(&mut OsRng);
        ChaCha20Poly1305::new(&key)
    });

    let encryption_key = ENCRYPTION_KEY.get_or_init(
        || ChaCha20Poly1305::generate_nonce(&mut OsRng), // 96-bits; unique per message
    );

    let ciphertext = cipher
        .encrypt(&encryption_key, something.as_slice())
        .unwrap();

    let mut writer = brotli::CompressorWriter::new(Vec::new(), 4096, 11, 22);
    writer.write_all(ciphertext.as_slice())?;
    let something_diff_2 = writer.into_inner();

    let something_diff_3 =
        base64::engine::general_purpose::STANDARD_NO_PAD.encode(something_diff_2.clone());

    println!("output (original): {}", some_content.as_bytes().len());
    println!("output (first compression): {}", something.len());
    println!("output (encryption): {}", ciphertext.len());
    println!("output (second compression): {:?}", something_diff_2.len());
    println!("output (base64): {}", something_diff_3.len());

    Ok(())
}
```

**Output of the sample code**

``` shell
output (original): 656
output (first compression): 304
output (encryption): 320
output (second compression): 324
output (base64): 459
```

**dependencies section of the Cargo.toml file of the sample code**

``` toml
[dependencies]
brotli = { version = "3.4.0", default-features = false, features=["std"] }
chacha20poly1305={version="0.10.1", default-features=false, features=["alloc","getrandom"]}
chacha20 = {version="0.9.1", default-features=false}
base64 = {version="0.21.5", default-features=false, features=["std"]}
```

evanyang1:
## What does this PR do?

This PR provides the compression and encryption of cached search results by providing four conditionally compiled features in rust, which looks something like this:
 
- `Default`: No compression and encryption
- `Compress-cache-results:`  Uses compression algorithm like `brotli` (which is considered good for text based compression)
- `Encrypt-cache-results:` Uses an encryption algorithm like `chacha20poly1305` (which is considered very strong encryption and is widely adopted in the `rust` programming language). Also, the generated encryption should change on each restart of the `websurfx` server, which help against attacks and improve the privacy of the user. 
-  `cec-cache-results:` Uses both compression and encryption to first compress the results then encrypt it and then compress it back again to get the advantages of both encryption and compression. 

This change is essential as by providing this feature:
- By compressing the cached search results, the search engine would consume less amount of memory required for storing cache results, as well as improve the speed of transferring cache results over the internet. The size of the cache results become smaller after compression. So, if the cache results were to be stored on a Redis server provided by a VPS provider.
- By encrypting cache results, this can help improve user privacy by not allowing the Redis VPS provider to know what is being cached on their Redis server. Though, the cache results size become larger after encryption.
- By using both the encryption and compression with `cec-cache-results` feature, we provide the advantages of both compression and encryption by first compressing the results and the encrypting it. Then compressing it again so that the size of the cache result remains nearly the same as it was after the first compression. 

It can be tested by installing and running `Websurfx` as mentioned in the `docs` and on the `readme` and by launching the browser and thoroughly testing. By using a memory profiler to check whether the app consumes less memory when the cached search results are compressed with the default non-compressed cached search results. Also, checking whether the results are getting encrypted when the `Encrypted-cache-results` features is enabled.  

## Author's checklist

- [x] provide the three conditionally compilation features
- [x] implement code for the `Compress-cache-results` feature.
- [x] implement code for the `Encrypted-cache-results` feature.
- [x] implement code for the `cec-cache-results` feature.
- [x] bump the app version to `v1.9.0`

## Related issues

Closes #416


Cargo.toml
```diff
@@ -38,6 +38,8 @@ mimalloc = { version = "0.1.38", default-features = false }
 async-once-cell = {version="0.5.3", default-features=false}
 actix-governor = {version="0.5.0", default-features=false}
 mini-moka = { version="0.10", optional = true, default-features=false, features=["sync"]}
+brotli = { version = "3.4.0", default-features = false, features=["std"] }
+chacha20poly1305={version="0.10.1", default-features=false, features=["alloc","getrandom"]}
```
```suggestion
brotli = { version = "3.4.0", default-features = false, features=["std"], optional=true}
chacha20poly1305={version="0.10.1", default-features=false, features=["alloc","getrandom"], optional=true}
```

I would suggest adding the `optional=true` after both this crates because this will be required when adding the feature/conditional compiling feature approach in the project. :slightly_smiling_face: 

For a guide on how to provide conditional compilation approach in rust, I would suggest watching this video: 

https://piped.video/watch?v=QkHhwYJaP0U

neon-mmd (changes requested):
Thanks :heart: for working on this PR. We really appreciate it :+1: Since this PR is still a draft, we would keep on reviewing it providing you with feedbacks for the changes you made time to time by suggesting changes/improvements to be made if required like the one below. :slightly_smiling_face:

Cargo.toml
evanyang1: Thank you for the feedback!

evanyang1:
To anyone, I would appreciate any hints and/or code scaffolding (or outline) for the encryption/compression part. I am new to Rust. Hopefully I would get faster at Rust and then do PRs faster!

src/cache/cacher.rs
```diff
@@ -57,6 +57,7 @@ pub trait Cacher: Send + Sync {
     /// failure.
     async fn cache_results(
         &mut self,
+        options: u8,
```
neon-mmd:
```suggestion
```

I think you are going with the code based approach of feature options. While the issue expects that you have to do it according to the rust's conditional compiling method of feature based approach. I think I would suggest you to watch the above-mentioned video tutorial on how to write code with rust's conditional feature for this. :slightly_smiling_face: 

Also, Here are some other resources on how to write code with conditional compiled feature:

- https://stackoverflow.com/questions/27632660/how-do-i-use-conditional-compilation-with-cfg-and-cargo
- https://betterprogramming.pub/compile-time-feature-flags-in-rust-why-how-when-129aada7d1b3
- https://stackoverflow.com/questions/60287822/can-i-create-my-own-conditional-compilation-attributes

Which can help you better understand and write code in this aspect. :slightly_smiling_face:

neon-mmd (changes requested):

src/cache/cacher.rs
evanyang1: I see the importance of cfg but I may need more hints :/

neon-mmd: > I see the importance of cfg but I may need more hints :/

@spencerjibz, I think you may have a better idea on this. Would you like to provide some hints to evanyang1 ? :slightly_smiling_face:

src/cache/cacher.rs
```diff
@@ -69,6 +70,42 @@ pub trait Cacher: Send + Sync {
...
+        };
+
+        let key = ChaCha20Poly1305::generate_key(&mut OsRng);
+        let cipher = ChaCha20Poly1305::new(&key);
+        let nonce = ChaCha20Poly1305::generate_nonce(&mut OsRng); // 96-bits; unique per message
+        bytes = cipher
+            .encrypt(&nonce, bytes.as_ref())
+            .map_err(|_| CacheError::EncryptionError)?;
+        Ok(bytes)
+    }
```
neon-mmd:
```suggestion
    #[cfg(any(
        feature = "compress-cache-results",
        feature = "encrypted-cache-results",
        feature = "cec-cache-results"
    ))]

    /// A helper function that returns either compressed or encryption results
    /// Feature flags are required  for this to work
    async fn compress_encrypt_results(
        &mut self,
        search_results: &str,
        url: &str,
    ) -> Result<Vec<u8>, Report<CacheError>> {
        let mut bytes = search_results.as_bytes();
        
        #[cfg(feature = "compress-cache-results")]
        {
	        use std::io::Write;
	        let mut writer = brotli::CompressorWriter::new(Vec::new(), 4096, 11, 22);
	        writer.write_all(&bytes);
	        bytes = writer.into_inner();
	        Ok(bytes)
        }
        #[cfg(feature = "encrypt-cache-results")]
        {
	        use chacha20poly1305::{
	            aead::{Aead, AeadCore, KeyInit, OsRng},
	            ChaCha20Poly1305,
	        };
	
		let cipher = CIPHER.get_or_init(|| {
		        let key = ChaCha20Poly1305::generate_key(&mut OsRng);
		        ChaCha20Poly1305::new(&key)
		});
		
		let encryption_key = ENCRYPTION_KEY.get_or_init(
		        || ChaCha20Poly1305::generate_nonce(&mut OsRng), // 96-bits; unique per message
		);
	
	        bytes = cipher
	            .encrypt(&encryption_key, bytes.as_ref())
	            .map_err(|_| CacheError::EncryptionError)?;
	        Ok(bytes)
        }
    }
```

I think you quite get the issue properly, like it was not clear. Sorry, I think we should have explained it in more depth, but we will take care of it next time. 

Actually, the issue talks about compressing/encrypting or both of the search results before putting them into the cache and the opposite after getting the results from the cache. So you will need to implement a function which can compress, encrypt and both of the search results in the trait and the decompress, decrypt and both of the search results in the trait. Then we can call them in the Redis and memory caching or fetching functions as needed. :slightly_smiling_face: 

As such, the implementation looks good, but I would suggest you to make the following changes: 
- Renaming the function to `compress_encrypt_results` because I think this will better match the work the function is doing.
- Splitting the encrypt and compress functions into two because this will make it easier to implement the third feature. As the third feature, will require you to implement a third function which can then call the other two function directly.
- Moving the code to generate cipher and encryption keys into `oncelock` which allows generating the encryption key and cipher once globally. Otherwise, generating each new encryption key on a new call of this function which will cause the results with the same urls to have encrypted with different keys, which can cause issues while fetching cached results and decrypting it. For example, the fetched cache results will never be decrypted because the currently generated new key will be different from the key with which the old results which were fetched from the cached were encrypted with.

Also, for the `oncelock` implementation you will need to add a new crate in the project which I would suggest adding it by running the following command:

```shell
cargo add chacha20 --no-default-features
``` 

Moreover, for the `oncelock` implementation you will need to declare a new static constant which I would suggest adding it something like this:

```rust
static CIPHER: OnceLock<
    ChaChaPoly1305<
        StreamCipherCoreWrapper<ChaChaCore<UInt<UInt<UInt<UInt<UTerm, B1>, B0>, B1>, B0>>>,
    >,
> = OnceLock::new();

static ENCRYPTION_KEY: OnceLock<GenericArray<u8, UInt<UInt<UInt<UInt<UTerm, B1>, B1>, B0>, B0>>> =
    OnceLock::new();
```

neon-mmd (changes requested):

src/cache/cacher.rs
```diff
@@ -69,6 +70,51 @@ pub trait Cacher: Send + Sync {
     fn hash_url(&self, url: &str) -> String {
         blake3::hash(url.as_bytes()).to_string()
     }
+    
+    /// A helper function that returns either compressed or encryption results
+    /// Feature flags are required  for this to work
```
neon-mmd:
```suggestion
    /// A helper function that returns either compressed or encryption results
    /// Feature flags are required  for this to work
    ///
    /// # Arguments 
    /// 
    /// * `search_results` - It takes the json search results string as an argument.
    /// * `url` - It takes the url of the queried pages as an argument.
    ///
    /// # Error
    /// Returns the compressed or encrypted bytes on success otherwise it returns a CacheError
    /// on failure.
```

Sorry for the delay in review. 
For the documentation around the functions, structs and enums, I would suggest following the rust convention for this. 

Here is the link to the rust convention on writing documentation around code:
- https://rust-lang.github.io/api-guidelines/documentation.html

Also, for more examples on how to write documentation around code, I would recommend reviewing other functions, structs and enums in the project. :slightly_smiling_face:

neon-mmd (changes requested):

Cargo.toml
```diff
@@ -38,6 +38,10 @@ mimalloc = { version = "0.1.38", default-features = false }
 async-once-cell = {version="0.5.3", default-features=false}
 actix-governor = {version="0.5.0", default-features=false}
 mini-moka = { version="0.10", optional = true, default-features=false, features=["sync"]}
+brotli = { version = "3.4.0", default-features = false, features=["std"], optional=true}
+chacha20poly1305={version="0.10.1", default-features=false, features=["alloc","getrandom"], optional=true}
+base64 = {version= "0.21.5", optional = true}
+chacha20 = {version = "0.9.1", optional = true}
```
```suggestion
chacha20 = {version="0.9.1", default-features=false, optional=true}
base64 = {version="0.21.5", default-features=false, features=["std"], optional=true}
```

I would suggest disabling the unneeded default-features as by disabling them it improves compilation times which in turns helps improve user experience. :slightly_smiling_face:

neon-mmd (changes requested):

neon-mmd:
I think this code already looks completed and can be moved from a draft PR to a PR, but before doing that if you wish to still refactor and optimize the code. Then I would suggest doing it and then moving this to a PR. :slightly_smiling_face:

spencerjibz:
> I think this code already looks completed and can be moved from a draft PR to a PR, but before doing that if you wish to still refactor and optimize the code. I suggest doing it and then moving this to a PR. 🙂

Thanks, I think it's pretty complete, I removed some unnecessary clone calls plus most of those functions didn't need to be async as well.
All helper methods were added to the trait implementation as provided methods, we don't repeat ourselves, a new cacher implementation would only need to call those methods before and after storing the values respectively.

neon-mmd:
> > I think this code already looks completed and can be moved from a draft PR to a PR, but before doing that if you wish to still refactor and optimize the code. I suggest doing it and then moving this to a PR. 🙂
> 
> Thanks, I think it's pretty complete, I removed some unnecessary clone calls plus most of those functions didn't need to be async as well. All helper methods were added to the trait implementation as provided methods, we don't repeat ourselves, a new cacher implementation would only need to call those methods before and after storing the values respectively.

You are welcome, but I think I would still suggest keeping the method `async` because I feel they might block the async thread for too long. What do you think? Will they? 

Also, yes, I did see, the methods in the trait implementation. Well, I think you have a done great job of deduplicating/abstracting code. I really appreciate it :+1: .

spencerjibz:
> > > I think this code already looks completed and can be moved from a draft PR to a PR, but before doing that if you wish to still refactor and optimize the code. I suggest doing it and then moving this to a PR. 🙂
> > 
> > 
> > Thanks, I think it's pretty complete, I removed some unnecessary clone calls plus most of those functions didn't need to be async as well. All helper methods were added to the trait implementation as provided methods, we don't repeat ourselves, a new cacher implementation would only need to call those methods before and after storing the values respectively.
> 
> You are welcome, but I think I would still suggest keeping the method `async` because I feel they might block the async thread for too long. What do you think? Will they?
 No, they won't block the thread, using async functions themselves calls with the overhead of pinning in on the heap (box::pin). It's fine cos the abstraction used inside of those functions isn't async too.   (basically using state-machine logic for sync functions).
> Also, yes, I did see, the methods in the trait implementation. Well, I think you have a done great job of deduplicating/abstracting code. I appreciate it 👍 .

No problem, we could just test the code, move the PR from daft and merge it if necessary.

spencerjibz:
Hi everyone,  I have cleaned all the bugs, tested every feature flag, and find some screenshots in our discord channel.  
### Tools for testing results.
-  Redisinsight or other desktop GUI Redis clients
-  [base64decode website](https://www.base64decode.org/)
-  Running the project with features flags enabled

src/bin/websurfx.rs
```diff
@@ -2,7 +2,7 @@
 //!
 //! This module contains the main function which handles the logging of the application to the
 //! stdout and handles the command line arguments provided and launches the `websurfx` server.
-
+#[allow(unused_imports)]
```
ibnz36: What's this for? You should use `#[cfg(feature = ...)]` for imports only used in features

src/cache/cacher.rs
```diff
@@ -69,6 +73,237 @@ pub trait Cacher: Send + Sync {
...
+    /// # Error
+    /// Returns a Vec of compressed or encrypted bytes on success otherwise it returns a CacheError
+    /// on failure.
+    fn pre_process_search_results(
+        &mut self,
+        search_results: &SearchResults,
+    ) -> Result<Vec<u8>, Report<CacheError>> {
+        #[allow(unused_mut)] // needs to be mutable when any of the features is enabled
+        let mut bytes: Vec<u8> = search_results.try_into()?;
+        #[cfg(feature = "compress-cache-results")]
```
ibnz36, line 223: `cfg_if` might be better here

src/cache/cacher.rs
```diff
@@ -69,6 +73,237 @@ pub trait Cacher: Send + Sync {
...
+    ///
+    /// # Error
+    /// Returns the SearchResults struct on success otherwise it returns a CacheError
+    /// on failure.
+
+    #[allow(unused_mut)] // needs to be mutable when any of the features is enabled
+    fn post_process_search_results(
+        &mut self,
+        mut bytes: Vec<u8>,
+    ) -> Result<SearchResults, Report<CacheError>> {
```
ibnz36, line 260: `cfg_if` might also be better here

src/cache/encryption.rs
```diff
@@ -0,0 +1,26 @@
...
+        typenum::{UInt, UTerm},
+        StreamCipherCoreWrapper,
+    },
+    ChaChaCore,
+};
+
+/// OUR CIPHER, only initialised once
+pub static CIPHER: OnceLock<
+    ChaChaPoly1305<
+        StreamCipherCoreWrapper<ChaChaCore<UInt<UInt<UInt<UInt<UTerm, B1>, B0>, B1>, B0>>>,
```
ibnz36: You might also want to specify here with a comment the reason behind this complex type

ibnz36:
I honestly think this is ready to merge, code style isn't strictly necessary for this to be merged and can be handled at any moment

src/cache/cacher.rs
spencerjibz: Adding an extra dependency is unnecessary here, especially an extra macro for compile time's sake. cfg-if is used as an optional dependency on another line.

src/bin/websurfx.rs
spencerjibz: Thanks, I'll add that change now.

src/cache/encryption.rs
spencerjibz: I think the reason here is very visible but that type is way too long. I'll add a comment there thanks.

spencerjibz (approved):

src/cache/encryption.rs
```diff
@@ -0,0 +1,26 @@
...
```
neon-mmd:
```suggestion
/// --- Add appropriate documentation for this type ---
type ChaChaCoreStreamCipher = ChaChaPoly1305<StreamCipherCoreWrapper<ChaChaCore<UInt<UInt<UInt<UInt<UTerm, B1>, B0>, B1>, B0>>>;

use chacha20poly1305::{
    consts::{B0, B1},
    ChaChaPoly1305,
};
use std::sync::OnceLock;

use chacha20::{
    cipher::{
        generic_array::GenericArray,
        typenum::{UInt, UTerm},
        StreamCipherCoreWrapper,
    },
    ChaChaCore,
};

/// OUR CIPHER, only initialised once
pub static CIPHER: OnceLock<ChaChaCoreStreamCipher>,
```

I think instead of allowing this Clippy lint, I would suggest instead defining a custom type and just replace it with it in the code. :slightly_smiling_face:

neon-mmd (changes requested):

src/cache/encryption.rs
```diff
@@ -1,4 +1,6 @@
-#![allow(clippy::type_complexity)] // the static types are too long, this cases no compilation problems though.
+/// --- Add appropriate documentation for this type ---
```
neon-mmd: One final minor suggestion I would like to make, I would suggest updating this with an appropriate documentation.

neon-mmd (changes requested):

src/cache/encryption.rs
spencerjibz: All of that is now getting resolved by the next commit

neon-mmd (approved):
Thanks :heart: again for working out such a wonderful PR. We really appreciate it :+1: Since everything looks good. We will merge the PR as soon as possible :slightly_smiling_face:

alamin655 (approved):

PR #451, opened 2023-12-21
🐛 Openssl runtime dependencies for `nix flake`

### LINKED ISSUE #448: 🐛 `Cargo test` failure when developing work on the app using `nix develop` due to missing `openssl` package
### Description

Provide a fix for the `cargo test` failure issue by providing a way to install the `openssl` package prior to running the command in the nix flake.

### Screenshots

_No response_

None

### Additional information

_No response_

jfvillablanca:
## What does this PR do?
- OpenSSL dependencies are usually readily available for non-Nix distros but not for NixOS users.

- Enables devs running NixOS to contribute

## Additional Notes

- I ran into trouble making this PR because `rusty-hook` CLI is not auto-installed when running `cargo test`. This is only specific to NixOS since global installs are not a thing. I am attempting to make a fix: package `rusty-hook` and include it in the flake.
- Another odd thing is that running `git commit --no-verify` doesn't work and still attempts to run `rusty-hook`. The hack I did is to comment out all hooks in `.git/hooks` that hinders: from `prepare-commit-msg` to `pre-push`. A real hassle. Any ideas why this is the case? 

## Related issues
Closes #448

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -41,6 +41,8 @@
             rustPackages.clippy
             rustc
             yamllint
+            openssl
+            pkg-config
           ];
           RUST_SRC_PATH = rustPlatform.rustLibSrc;
         };


jfvillablanca:
> * I am attempting to make a fix: package `rusty-hook` and include it in the flake.

I found the reason ([rusty-hook issue](https://github.com/swellaby/rusty-hook/issues/168)) why `rusty-hook` did not work even if the binary is packaged. The installation is tightly coupled to cargo, thus needing to be installed via cargo and cannot work when running as a standalone binary. I'll try to document the workaround and make a PR

neon-mmd:
> > * I am attempting to make a fix: package `rusty-hook` and include it in the flake.
> 
> I found the reason ([rusty-hook issue](https://github.com/swellaby/rusty-hook/issues/168)) why `rusty-hook` did not work even if the binary is packaged. The installation is tightly coupled to cargo, thus needing to be installed via cargo and cannot work when running as a standalone binary. I'll try to document the workaround and make a PR

Yes, you are right.  `rusty-hook` tends to be highly coupled to cargo I think maybe a fix for this would be to use something like `cargo install` to install the rusty-hook package without the need to run cargo test everytime to install it. I think for this we would suggest opening an issue. :slightly_smiling_face:

neon-mmd:
> From the additional notes:
> Another odd thing is that running git commit --no-verify doesn't work and still attempts to run rusty-hook. The hack I did is to comment out all hooks in .git/hooks that hinders: from prepare-commit-msg to pre-push. A real hassle. Any ideas why this is the case?

I think for this we will need to investigate the issue because on my development environment it seems to work without any issues like when I run `git commit --no-verify` it seems to work fine and it even doesn't `rusty-hook` . So again it could be distro specific issue. So again for this we would suggest opening an issue. :slightly_smiling_face:

neon-mmd (approved):
Thanks :heart: for working on such a wonderful PR. We really appreciate it :+1: Keep it up!! Since everything looks good. We will merge this PR right away :rocket: :slightly_smiling_face:

jfvillablanca:
> > > * I am attempting to make a fix: package `rusty-hook` and include it in the flake.
> > 
> > 
> > I found the reason ([rusty-hook issue](https://github.com/swellaby/rusty-hook/issues/168)) why `rusty-hook` did not work even if the binary is packaged. The installation is tightly coupled to cargo, thus needing to be installed via cargo and cannot work when running as a standalone binary. I'll try to document the workaround and make a PR
> 
> Yes, you are right. `rusty-hook` tends to be highly coupled to cargo I think maybe a fix for this would be to use something like `cargo install` to install the rusty-hook package without the need to run cargo test everytime to install it. I think for this we would suggest opening an issue. 🙂

I was able to make a "hacky" workaround. I think Nix doesn't "allow" the installation of `rusty-hook` on the first `cargo test` so you have to manually `cargo install rusty-hook` and add this:
```nix
mkShell {
  # ...

  # any arbitrary path for impure installations
  shellHook = ''
    export PATH="$PATH:$HOME/.cargo/bin"
  '';
}
```
to the flake.nix.
Until the point when upstream `rusty-hook` decouples itself from cargo, I think this is the way for now.

PR #454, opened 2023-12-23
:memo:  Developer environment setup on NixOS 

### LINKED ISSUE #453: 📝 NixOS development environment setup
### Description

- The current instructions for setting up a developer environment are suboptimal and are using assumptions that does not work in NixOS.
- For example: `docker`, `rust-analyzer`, `rustfmt` can be included inside `mkShell.buildInputs`. `nodejs` also needs to be added to install dev dependencies such as `stylelint-config-standard` and `postcss-lit` which are used by pre-commit hooks.

### Screenshots

_No response_

Yes

### Additional information

_No response_

jfvillablanca:
## What does this PR do?

- Update the flake to add more dev environment dependencies
- Update the docs on the NixOS section

- To increase the reproducibility of the dev environment for contributors using NixOS

- Any contributor that uses NixOS, can you please replicate the instructions in the docs on a fresh `git clone` of the repo to check if the dev environment is properly setup and if the instructions are clear enough. 

## Related issues
Closes #453

### DIFF
--- a/docs/developing.md
+++ b/docs/developing.md
@@ -212,12 +212,6 @@ This section covers how to setup the project for development using the `NixOS de
 Before you start working on the project. You will need the following packages installed on your system:
 
 - `Git` installed on your system. The installation instructions for this can be found [here](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git).
-- Finally, The latest version of `Docker` is installed on your system which will be used to avoid introducing unexpected issues when working on the project. The installation instructions for this can be found [here](https://docs.docker.com/engine/install/).
-
-> Optionally, On `NixOS` the above-mentioned required packages except for `stylelint` and `cargo-watch` could also be installed by following the link to the installation instructions provided below:
->
-> - `Git`: https://search.nixos.org/packages?channel=23.05&show=git&from=0&size=50&sort=relevance&type=packages&query=git
-> - `Docker`: https://search.nixos.org/packages?channel=23.05&show=docker&from=0&size=50&sort=relevance&type=packages&query=docker
 
 #### Setting up Pre-commit Checks
 
@@ -234,7 +228,11 @@ Then run the following command to setup the `NixOS dev-shell`:
 nix develop
 ```
 
-Once you have finished running the above commands then run the following command to setup the `pre-commit` checks:
+> Alternatively, you can use `nix-direnv` to simplify entering into the `nix-shell` but the setup is beyond the scope here. Read more about it here: [nix-direnv](https://github.com/nix-community/nix-direnv)
+
+This will add `docker`, `cargo-watch`, and other dev environment essentials to your `nix-shell` so you don't have to install all of them imperatively.
+
+After finishing the commands above, run the following command to setup the `pre-commit` checks:
 
 ```shell
 cargo test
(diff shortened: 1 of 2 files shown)


docs/developing.md
```diff
@@ -244,14 +242,22 @@ By running the above-mentioned command, it will automatically set up all the pre
 
 #### Post Setup Requirements
 
-After you have done setting up pre-commit checks, then you may need to fulfill a few more requirements to finish setting up the development environment with `NixOS dev-shell`. These include:
+Inside `flake.nix`, the `$HOME/.cargo/bin` is added to the `PATH`. The git hook manager for this project is `rusty-hook`, and (unfortunately) it has to be installed via `cargo` and can't be neatly packaged up inside the flake ([see issue](https://github.com/swellaby/rusty-hook/issues/168)). This was automatically installed when you ran `cargo test`.
```
neon-mmd:
```suggestion
```

I think I would suggest not including this part because I think the documentation should contain instruction which should be straight to the point. Also, specially developer documentation should be very straight to the point otherwise it gives a very bad impression.

For example:

here is a developer documentation from some project:

- https://github.com/neovim/neovim/blob/master/CONTRIBUTING.md
- https://github.com/Lissy93/dashy/blob/master/docs/developing.md
- https://github.com/libp2p/rust-libp2p/blob/master/docs/coding-guidelines.md

You can see that the developer documentation tends to be pretty straight to the point, with the explanations as required. That's why I would suggest not including it. :slightly_smiling_face:

docs/developing.md
```diff
@@ -244,14 +242,22 @@ By running the above-mentioned command, it will automatically set up all the pre
...
 
-- `Cargo-watch` installed on your system which will allow you to auto-build the project when any checks occur in the source code files in the codebase (`websurfx` directory). Before you install `cargo-watch` on your system, make sure you have `cargo` installed on your system. To install `cargo-watch` run the following command:
+Run `git commit`. This should take a while and the pre-commit hook will get rejected and we will explain why.
```
neon-mmd: I think I would suggest reframing this part a little because it seems to be repeated two times, which is not good. Also, as mentioned in the first code review, I think I would suggest making it a little straight forward, something like this. :slightly_smiling_face: :

> Finally, you will need to install the following development dependencies to get the pre-commit fully setup....

neon-mmd (changes requested):
Thanks :heart: for working on this PR. We really appreciate it :+1: But before we merge this PR, we would like to suggest a few changes :slightly_smiling_face: .

docs/developing.md
jfvillablanca: Thank you for the feedback, will get to work on this. 
Question:
- Do you prefer having new commits on the revision,
- or, do you want a "clean" git history where I would force push my revision,
- or, do you prefer that I write my revision here as comments so that you can give me some feedback before making a new commit?

neon-mmd: We would suggest having revisions as new commits that would be better because at the end of the day all PRs are squashed and merged so the git history always remains clean. :slightly_smiling_face:

neon-mmd (approved):
Thanks :heart: again for working on this. We really appreciate your work :+1: Since everything looks good. We will merge this PR right away :rocket: :slightly_smiling_face: .

PR #467, opened 2023-12-29
:recycle: Start tracking page numbers from 0 instead of 1 in the backend

### LINKED ISSUE #459: :recycle:  The page number for upstream search queries should start at 0
### Description

The back-end starts the page number count at 1 and has different code for upstream search engines to handle page 1 and 0. Changing the back end to start at 0 instead will help keep the design and implementations consistent.

### Screenshots

_No response_

Yes

### Additional information

_No response_

greengeckoz:
## What does this PR do?

Modified server::routes::search.rs to subtract one from the page number that's displayed to the user. So if the user sees they're on page 1, the backend will instead treat it as page 0 to be consistent with other search engines and programming lists.

This will help reduce confusion by allowing the different implementations of search engines to be more consistent with one another.

Test that the search results appear as expected. Right now the UI will let you go down from page 1 to  page 0.

## Related issues

Potentially closes #459

### DIFF
--- a/src/server/routes/search.rs
+++ b/src/server/routes/search.rs
@@ -59,11 +59,11 @@ pub async fn search(
                 )
             };
 
-            // .max(1) makes sure that the page > 0.
-            let page = params.page.unwrap_or(1).max(1);
+            // .max(1) makes sure that the page >= 0.
+            let page = params.page.unwrap_or(1).max(1) - 1;
 
             let (_, results, _) = join!(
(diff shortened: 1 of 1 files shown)


neon-mmd (approved):
Thanks :heart: for working on such a wonderful PR. We really appreciate it :+1: Keep it up!! Since everything looks good. We will merge this PR right away :rocket: :slightly_smiling_face: .

PR #470, opened 2023-12-31
:bug: Different cache keys when different search engines are selected

### LINKED ISSUE #463: 🐛 Results from different search engines get cached as the same key
### Description

If the user changes their settings to use different search engines and repeats a search they already made, the cache will incorrectly return the old results instead of doing a new search with the new selected engines. 

The cache uses the URL as a key to lookup and store the search results, so the URL will have to be modified to reflect which search engines were used.

### Screenshots

_No response_

None

### Additional information

_No response_

greengeckoz:
## What does this PR do?

Modified server::routes::search.rs to add the search engines used in the search and in the browser's cookie.

Now repeat searches will return different results if the engines used are changed. While still returning cached results if no settings were changed.

Make sure the results are still returned as expected. The results should be the quick and cached while searching  for the same query with no changes to the engines used. 

## Related issues

Closes #463


src/server/routes/search.rs
```diff
@@ -123,11 +123,34 @@ async fn results(
...
+        let mut engines: Vec<String> = config
+            .upstream_search_engines
+            .iter()
+            .filter(|map| *map.1)
+            .map(|map| String::from(map.0))
+            .collect();
+
+        engines.sort();
+        cache_key = cache_key + &(engines.join(""));
+    }
```
neon-mmd:
```suggestion
 
```

I think I would suggest instead of having a duplicate cookie handling code, you can add the search engine addition code for the URL search parameter in the already existing cookie handling code which can be found here:

https://github.com/neon-mmd/websurfx/blob/5b4864424a3f740f13749185505385ed66f46644/src/server/routes/search.rs#L156-L212  

Also, instead of adding the search engines with ("") I would suggest adding them comma separated (",") which I think would be better.

Moreover, instead of using (+) for concatenation, I would recommend using the `format!` macro, which tends to be more idiomatic and rusty. :slightly_smiling_face:

neon-mmd (changes requested):
Thanks :heart: for working on this PR. We really appreciate it :+1: but before merge this PR, we would like to make a few suggestions. :slightly_smiling_face:

src/server/routes/search.rs
greengeckoz: I can fix the separator and concatenation but I'm confused by what you mean for the cookie handling code? Do you want me to move the use of some of the cookie's variables `engines` to before the cache key is built and modified and then reuse it later on when it was used before?

neon-mmd: > I can fix the separator and concatenation but I'm confused by what you mean for the cookie handling code? Do you want me to move the use of some of the cookie's variables `engines` to before the cache key is built and modified and then reuse it later on when it was used before?

No I am actually suggesting to move the code related to `engines` and everything within the already existing cookie handling code present in the `results` function. I hope you got my point :slightly_smiling_face: .

greengeckoz: I pushed another commit where I moved the checking/parsing of cookie value and the config file to before the cache_key is built. I then reuse that parsed data later on if/when the aggregate function uses them to create the list of EngineHandlers for the search.

neon-mmd: > I pushed another commit where I moved the checking/parsing of cookie value and the config file to before the cache_key is built. I then reuse that parsed data later on if/when the aggregate function uses them to create the list of EngineHandlers for the search.

Sorry, for the late reply.

I think you still didn't get my point :smile: . I think now I will provide you with a sample code which might give you a better idea of what I am suggesting. So I would suggest taking a look. :slightly_smiling_face: 

**Sample Code**
```diff
/// Fetches the results for a query and page. It First checks the redis cache, if that
/// fails it gets proper results by requesting from the upstream search engines.
///
/// # Arguments
///
/// * `url` - It takes the url of the current page that requested the search results for a
/// particular search query.
/// * `config` - It takes a parsed config struct.
/// * `query` - It takes the page number as u32 value.
/// * `req` - It takes the `HttpRequest` struct as a value.
///
/// # Error
///
/// It returns the `SearchResults` struct if the search results could be successfully fetched from
/// the cache or from the upstream search engines otherwise it returns an appropriate error.
async fn results(
    config: &Config,
    cache: &web::Data<SharedCache>,
    query: &str,
    page: u32,
    req: HttpRequest,
    safe_search: &Option<u8>,
) -> Result<SearchResults, Box<dyn std::error::Error>> {
    // eagerly parse cookie value to evaluate safe search level
    let cookie_value = req.cookie("appCookie");

    let cookie_value: Option<Cookie<'_>> = cookie_value
        .as_ref()
        .and_then(|cv| serde_json::from_str(cv.name_value().1).ok());

    let safe_search_level = get_safesearch_level(
        safe_search,
        &cookie_value.as_ref().map(|cv| cv.safe_search_level),
        config.safe_search,
    );

+   let mut cache_key = format!(
        "http://{}:{}/search?q={}&page={}&safesearch={}",
        config.binding_ip, config.port, query, page, safe_search_level
    );

    // fetch the cached results json.
    let cached_results = cache.cached_results(&cache_key).await;
    // check if fetched cache results was indeed fetched or it was an error and if so
    // handle the data accordingly.
    match cached_results {
        Ok(results) => Ok(results),
        Err(_) => {
            if safe_search_level == 4 {
                let mut results: SearchResults = SearchResults::default();

                let flag: bool =
                    !is_match_from_filter_list(file_path(FileType::BlockList)?, query)?;
                // Return early when query contains disallowed words,
                if flag {
                    results.set_disallowed();
                    cache.cache_results(&results, &cache_key).await?;
                    results.set_safe_search_level(safe_search_level);
                    return Ok(results);
                }
            }

            // check if the cookie value is empty or not if it is empty then use the
            // default selected upstream search engines from the config file otherwise
            // parse the non-empty cookie and grab the user selected engines from the
            // UI and use that.
            let mut results: SearchResults = match cookie_value {
                Some(cookie_value) => {
                    let engines: Vec<EngineHandler> = cookie_value
                        .engines
                        .iter()
                        .filter_map(|name| EngineHandler::new(name).ok())
                        .collect();

+                   let mut engine_temp = cookie_value.engines;
+                   engine_temp.sort();

+                   cache_key = format!("{cache_key}&engines={}", engine_temp.join(","));

+                   drop(engine_temp);

                    match engines.is_empty() {
                        false => {
                            aggregate(
                                query,
                                page,
                                config.aggregator.random_delay,
                                config.debug,
                                &engines,
                                config.request_timeout,
                                safe_search_level,
                            )
                            .await?
                        }
                        true => {
                            let mut search_results = SearchResults::default();
                            search_results.set_no_engines_selected();
                            search_results
                        }
                    }
                }
+               None => {
+                   cache_key = format!(
+                       "{cache_key}&engines={}",
+                       &config
+                           .upstream_search_engines
+                           .clone()
+                           .into_keys()
+                           .collect::<Vec<String>>()
+                           .join(",")
+                   );

+                   aggregate(
+                       query,
+                       page,
+                       config.aggregator.random_delay,
+                       config.debug,
+                       &config
+                           .upstream_search_engines
+                           .clone()
+                           .into_iter()
+                           .filter_map(|(key, value)| value.then_some(key))
+                           .map(|engine| EngineHandler::new(&engine))
+                           .collect::<Result<Vec<EngineHandler>, error_stack::Report<EngineError>>>(
+                           )?,
+                       config.request_timeout,
+                       safe_search_level,
+                   )
+                   .await?
+               }
+           };
            if results.engine_errors_info().is_empty()
                && results.results().is_empty()
                && !results.no_engines_selected()
            {
                results.set_filtered();
            }
            cache.cache_results(&results, &cache_key).await?;
            results.set_safe_search_level(safe_search_level);
            Ok(results)
        }
    }
} 
```

greengeckoz: This example would have a bug however, the cache_key variable is used almost immediately to get the cache_results,  therefore, if we try to retrieve results that don't have the modifications you want to do later on it wont return the same result. That's why we need to do all the modifications to cache_key and have its final value before 
`let cached_results = cache.cached_results(&cache_key).await;`

greengeckoz: Basically, in this example code, the key would be saved with the engine data, but wouldn't be loaded/checked with a key that has the engine data. So we can never read any of the data that's saved to the cache.

neon-mmd: > This example would have a bug however, the cache_key variable is used almost immediately to get the cache_results, therefore, if we try to retrieve results that don't have the modifications you want to do later on it wont return the same result. That's why we need to do all the modifications to cache_key and have its final value before `let cached_results = cache.cached_results(&cache_key).await;`

Yes, you are right, I almost forgot about that error, but now I have made several updates to the sample code which I have attached down below. So I would suggest having a look:

**src/modals/server_modals.rs**
```diff
 //! This module provides the models to parse cookies and search parameters from the search
 //! engine website.
 use serde::Deserialize;
 
+use super::parser_models::Style;
+
 /// A named struct which deserializes all the user provided search parameters and stores them.
 #[derive(Deserialize)]
 pub struct SearchParams {
     /// It stores the search parameter option `q` (or query in simple words)
     /// of the search url.
     pub q: Option<String>,
     /// It stores the search parameter `page` (or pageno in simple words)
     /// of the search url.
     pub page: Option<u32>,
     /// It stores the search parameter `safesearch` (or safe search level in simple words) of the
     /// search url.
     pub safesearch: Option<u8>,
 }
 
 /// A named struct which is used to deserialize the cookies fetched from the client side.
 #[allow(dead_code)]
 #[derive(Deserialize)]
 pub struct Cookie<'a> {
     /// It stores the theme name used in the website.
     pub theme: &'a str,
     /// It stores the colorscheme name used for the website theme.
     pub colorscheme: &'a str,
     /// It stores the user selected upstream search engines selected from the UI.
-    pub engines: Vec<&'a str>,
+    pub engines: Vec<String>,
     /// It stores the user selected safe search level from the UI.
     pub safe_search_level: u8,
 }
+
+impl<'a> Cookie<'a> {
+    ///
+    pub fn new(style: &'a Style, safe_search_level: u8, mut engines: Vec<String>) -> Self {
+        //
+        engines.sort();
+
+        Self {
+            theme: &style.theme,
+            colorscheme: &style.colorscheme,
+            engines,
+            safe_search_level,
+        }
+    }
+}
```

**src/server/routes/search.rs**
```diff
//! This module handles the search route of the search engine website.
 
 use crate::{
     cache::cacher::SharedCache,
     config::parser::Config,
     handler::{file_path, FileType},
     models::{
         aggregation_models::SearchResults,
         engine_models::{EngineError, EngineHandler},
         server_models::{Cookie, SearchParams},
     },
     results::aggregator::aggregate,
 };
 use actix_web::{get, http::header::ContentType, web, HttpRequest, HttpResponse};
 use regex::Regex;
 use std::{
     fs::File,
     io::{BufRead, BufReader, Read},
 };
 use tokio::join;
 
 /// Handles the route of search page of the `websurfx` meta search engine website and it takes
 /// two search url parameters `q` and `page` where `page` parameter is optional.
 ///
 /// # Example
 ///
 /// ```bash
 /// curl "http://127.0.0.1:8080/search?q=sweden&page=1"
 /// ```
 ///
 /// Or
 ///
 /// ```bash
 /// curl "http://127.0.0.1:8080/search?q=sweden"
 /// ```
 #[get("/search")]
 pub async fn search(
     req: HttpRequest,
     config: web::Data<Config>,
     cache: web::Data<SharedCache>,
 ) -> Result<HttpResponse, Box<dyn std::error::Error>> {
     let params = web::Query::<SearchParams>::from_query(req.query_string())?;
     match &params.q {
         Some(query) => {
             if query.trim().is_empty() {
                 return Ok(HttpResponse::TemporaryRedirect()
                     .insert_header(("location", "/"))
                     .finish());
             }
 
+            // eagerly parse cookie value to evaluate safe search level
+            let cookie_value = req.cookie("appCookie");
+
+            let cookie_value_or_config_values: Cookie<'_> = match cookie_value {
+                Some(ref cookie) => serde_json::from_str(cookie.value())?,
+                None => Cookie::new(
+                    &config.style,
+                    config.safe_search,
+                    config
+                        .upstream_search_engines
+                        .clone()
+                        .into_iter()
+                        .filter_map(|(key, value)| value.then_some(key))
+                        .collect(),
+                ),
+            };
+
+            let safe_search_level = get_safesearch_level(
+                params.safesearch.as_ref(),
+                cookie_value_or_config_values.safe_search_level,
+                config.safe_search,
+            );
+
             let get_results = |page| {
                 results(
                     &config,
                     &cache,
                     query,
                     page,
-                    req.clone(),
-                    &params.safesearch,
+                    &cookie_value_or_config_values,
+                    format!(
+                        "http://{}:{}/search?q={}&page={}&safesearch={}&engines={}",
+                        config.binding_ip,
+                        config.port,
+                        query,
+                        page,
+                        safe_search_level,
+                        cookie_value_or_config_values.engines.join(",")
+                    ),
                 )
             };
 
             // .max(1) makes sure that the page >= 0.
             let page = params.page.unwrap_or(1).max(1) - 1;
 
             let (_, results, _) = join!(
                 get_results(page.saturating_sub(1)),
                 get_results(page),
                 get_results(page + 1)
             );
 
             Ok(HttpResponse::Ok().content_type(ContentType::html()).body(
                 crate::templates::views::search::search(
                     &config.style.colorscheme,
                     &config.style.theme,
                     &config.style.animation,
                     query,
                     &results?,
                 )
                 .0,
             ))
         }
         None => Ok(HttpResponse::TemporaryRedirect()
             .insert_header(("location", "/"))
             .finish()),
     }
 }
 
 /// Fetches the results for a query and page. It First checks the redis cache, if that
 /// fails it gets proper results by requesting from the upstream search engines.
 ///
 /// # Arguments
 ///
 /// * `url` - It takes the url of the current page that requested the search results for a
 /// particular search query.
 /// * `config` - It takes a parsed config struct.
 /// * `query` - It takes the page number as u32 value.
 /// * `req` - It takes the `HttpRequest` struct as a value.
 ///
 /// # Error
 ///
 /// It returns the `SearchResults` struct if the search results could be successfully fetched from
 /// the cache or from the upstream search engines otherwise it returns an appropriate error.
 async fn results(
     config: &Config,
     cache: &web::Data<SharedCache>,
     query: &str,
     page: u32,
-    req: HttpRequest,
-    safe_search: &Option<u8>,
+    cookie_or_config: &Cookie<'_>,
+    cache_key: String,
 ) -> Result<SearchResults, Box<dyn std::error::Error>> {
-    // eagerly parse cookie value to evaluate safe search level
-    let cookie_value = req.cookie("appCookie");
-
-    let cookie_value: Option<Cookie<'_>> = cookie_value
-        .as_ref()
-        .and_then(|cv| serde_json::from_str(cv.name_value().1).ok());
-
-    let safe_search_level = get_safesearch_level(
-        safe_search,
-        &cookie_value.as_ref().map(|cv| cv.safe_search_level),
-        config.safe_search,
-    );
-
-    let cache_key = format!(
-        "http://{}:{}/search?q={}&page={}&safesearch={}",
-        config.binding_ip, config.port, query, page, safe_search_level
-    );
-
     // fetch the cached results json.
     let cached_results = cache.cached_results(&cache_key).await;
     // check if fetched cache results was indeed fetched or it was an error and if so
     // handle the data accordingly.
     match cached_results {
         Ok(results) => Ok(results),
         Err(_) => {
-            if safe_search_level == 4 {
+            if cookie_or_config.safe_search_level == 4 {
                 let mut results: SearchResults = SearchResults::default();
 
                 let flag: bool =
                     !is_match_from_filter_list(file_path(FileType::BlockList)?, query)?;
                 // Return early when query contains disallowed words,
                 if flag {
                     results.set_disallowed();
                     cache.cache_results(&results, &cache_key).await?;
-                    results.set_safe_search_level(safe_search_level);
+                    results.set_safe_search_level(cookie_or_config.safe_search_level);
                     return Ok(results);
                 }
             }
 
             // check if the cookie value is empty or not if it is empty then use the
             // default selected upstream search engines from the config file otherwise
             // parse the non-empty cookie and grab the user selected engines from the
             // UI and use that.
-            let mut results: SearchResults = match cookie_value {
-                Some(cookie_value) => {
-                    let engines: Vec<EngineHandler> = cookie_value
-                        .engines
-                        .iter()
-                        .filter_map(|name| EngineHandler::new(name).ok())
-                        .collect();
-
-                    match engines.is_empty() {
-                        false => {
-                            aggregate(
-                                query,
-                                page,
-                                config.aggregator.random_delay,
-                                config.debug,
-                                &engines,
-                                config.request_timeout,
-                                safe_search_level,
-                            )
-                            .await?
-                        }
-                        true => {
-                            let mut search_results = SearchResults::default();
-                            search_results.set_no_engines_selected();
-                            search_results
-                        }
-                    }
+            let engines: Vec<EngineHandler> = cookie_or_config
+                .engines
+                .iter()
+                .map(|engine| EngineHandler::new(engine))
+                .collect::<Result<Vec<EngineHandler>, error_stack::Report<EngineError>>>()?;
+
+            let mut results = match engines.is_empty() {
+                false => {
+                    aggregate(
+                        query,
+                        page,
+                        config.aggregator.random_delay,
+                        config.debug,
+                        &engines,
+                        config.request_timeout,
+                        cookie_or_config.safe_search_level,
+                    )
+                    .await?
+                }
+                true => {
+                    let mut search_results = SearchResults::default();
+                    search_results.set_no_engines_selected();
+                    search_results
                 }
-                None => aggregate(
-                    query,
-                    page,
-                    config.aggregator.random_delay,
-                    config.debug,
-                    &config
-                        .upstream_search_engines
-                        .clone()
-                        .into_iter()
-                        .filter_map(|(key, value)| value.then_some(key))
-                        .map(|engine| EngineHandler::new(&engine))
-                        .collect::<Result<Vec<EngineHandler>, error_stack::Report<EngineError>>>(
-                        )?,
-                    config.request_timeout,
-                    safe_search_level,
-                )
-                .await?,
             };
             if results.engine_errors_info().is_empty()
                 && results.results().is_empty()
                 && !results.no_engines_selected()
             {
                 results.set_filtered();
             }
             cache.cache_results(&results, &cache_key).await?;
-            results.set_safe_search_level(safe_search_level);
+            results.set_safe_search_level(cookie_or_config.safe_search_level);
             Ok(results)
         }
     }
 }
 
 /// A helper function which checks whether the search query contains any keywords which should be
 /// disallowed/allowed based on the regex based rules present in the blocklist and allowlist files.
 ///
 /// # Arguments
 ///
 /// * `file_path` - It takes the file path of the list as the argument.
 /// * `query` - It takes the search query to be checked against the list as an argument.
 ///
 /// # Error
 ///
 /// Returns a bool indicating whether the results were found in the list or not on success
 /// otherwise returns a standard error type on a failure.
 fn is_match_from_filter_list(
     file_path: &str,
     query: &str,
 ) -> Result<bool, Box<dyn std::error::Error>> {
     let mut reader = BufReader::new(File::open(file_path)?);
     for line in reader.by_ref().lines() {
         let re = Regex::new(&line?)?;
         if re.is_match(query) {
             return Ok(true);
         }
     }
 
     Ok(false)
 }
 
 /// A helper function which returns the safe search level based on the url params
 /// and cookie value.
 ///
 /// # Argurments
 ///
 /// * `safe_search` - Safe search level from the url.
 /// * `cookie` - User's cookie
 /// * `default` - Safe search level to fall back to
-fn get_safesearch_level(safe_search: &Option<u8>, cookie: &Option<u8>, default: u8) -> u8 {
-    match safe_search {
+fn get_safesearch_level(
+    config_safe_search: Option<&u8>,
+    cookie_safe_search_level: u8,
+    default: u8,
+) -> u8 {
+    match config_safe_search {
         Some(ss) => {
             if *ss >= 3 {
                 default
             } else {
                 *ss
             }
         }
-        None => cookie.unwrap_or(default),
+        None => cookie_safe_search_level,
     }
 }
```

Also, note the sample code that I have attached this time affects code from multiple files which include `search.rs` and `server_modals` file. Additionally, I have not added/updated comments/documentation in some area because it is a sample code and may not be perfect and may require refactoring or changing the code in some area. So I would suggest adding/updating comments/documentation to the code if necesssary from your side. :slightly_smiling_face:

greengeckoz: Yeah this should work, I was trying to make logic changes without changing any function parameters or definitions. What if we, in an attempt to make it a bit more rusty, create a sort of result argument struct that has a default implementation using the config file and can be changed with the engines found in the cookie. We can then pass the struct as a singe parameter for the result function to use. 

We might need to make a change like that later to make sure everything the result function requests is Send and 'static or otherwise works with tokio::spawn.

neon-mmd: > Yeah this should work, I was trying to make logic changes without changing any function parameters or definitions. What if we, in an attempt to make it a bit more rusty, create a sort of result argument struct that has a default implementation using the config file and can be changed with the engines found in the cookie. We can then pass the struct as a singe parameter for the result function to use.
> 
> We might need to make a change like that later to make sure everything the result function requests is Send and 'static or otherwise works with tokio::spawn.

Yes, that would be better, I think we can have those changes in another PR after this. What do you say? :slightly_smiling_face: .

src/server/routes/search.rs
```diff
@@ -236,24 +228,3 @@ fn is_match_from_filter_list(
...
-        Some(ss) => {
-            if *ss >= 3 {
-                default
-            } else {
-                *ss
-            }
-        }
-        None => cookie.unwrap_or(default),
-    }
-}
```
neon-mmd: I would suggest keeping this function and using it in the code because this function basically takes the safe_search_level values from the config, search URL parameter and the cookie. If the config safe_search_level is 3 or 4 then it uses the config's safe_search_level. This disallows the user from overriding it by the cookie or the search URL parameters. Also, this function also, checks whether the search URL parameter was provided if not then it falls back to using the cookie's safe_search_level (if cookie was set) otherwise it falls back to using the config's safe_search_level. 

The reason behind having this safe_search_level handling function is to allow the user (server administrator) to provide an SFW only server where the search results do not contain any NSFW search results (content). I hope you got the idea behind having this function. :slightly_smiling_face: .

neon-mmd (changes requested):

src/server/routes/search.rs
```diff
@@ -151,51 +165,29 @@ async fn results(
...
-                            .await?
-                        }
-                        true => {
-                            let mut search_results = SearchResults::default();
-                            search_results.set_no_engines_selected();
-                            search_results
-                        }
-                    }
+            let mut results: SearchResults = match user_settings.engines.is_empty() {
+                false => {
```
ibnz36, line 167: Using a match statement for a boolean. You probably want to use `if`

src/models/server_models.rs
```diff
@@ -19,13 +21,26 @@ pub struct SearchParams {
 /// A named struct which is used to deserialize the cookies fetched from the client side.
 #[allow(dead_code)]
 #[derive(Deserialize)]
-pub struct Cookie<'a> {
+pub struct Cookie {
```
ibnz36: Is it really necessary to convert `&'a str` to `String`s? This might be worse for performance

src/server/routes/search.rs
neon-mmd: > Using a match statement for a boolean. You probably want to use `if`

Thanks :heart: for taking the time to provide your code revew. We really appreciate it :+1: . :slightly_smiling_face: 

Well, we prefer having `match` over `if` because for code consistency. So we would suggest leaving it with the `match` approach. :slightly_smiling_face: .

src/models/server_models.rs
greengeckoz: Yeah, I was getting annoying borrowed value doesn't live long enough when building a Cookie from other borrowed values.

src/server/routes/search.rs
greengeckoz: Yeah I thought my code already handled all the cases but I forgot about the safe search level from the params in the URL. But I noticed we were now handling the potential cookie and then config safe search levels twice. 

We can use a single line like this `if let Some(level) = &params.safesearch { search_settings.safe_search_level = *level; }` or a match statement if preferred. `match &params.safesearch {
                Some(level) => { search_settings.safe_search_level = *level; },
                None => {},
            }`

and then have the same result

greengeckoz: Could you explain why levels 3 and above fall back to the default in the URL? Is it because those levels aren't implemented correctly in some search engines?

Edit: To keep checking for that, we can modify get_safesearch_level itself to work like this

`search_settings.safe_search_level = get_safesearch_level(&params.safesearch, config.safe_search);`

```
fn get_safesearch_level(safe_search: &Option<u8>, default: u8) -> u8 {
    match safe_search {
        Some(ss) => {
            if *ss >= 3 {
                default
            } else {
                *ss
            }
        }
        None => default,
    }
}
```

neon-mmd: > Could you explain why levels 3 and above fall back to the default in the URL? Is it because those levels aren't implemented correctly in some search engines?
> 
> Edit: To keep checking for that, we can modify get_safesearch_level itself to work like this
> 
> `search_settings.safe_search_level = get_safesearch_level(&params.safesearch, config.safe_search);`
> 
> ```
> fn get_safesearch_level(safe_search: &Option<u8>, default: u8) -> u8 {
>     match safe_search {
>         Some(ss) => {
>             if *ss >= 3 {
>                 default
>             } else {
>                 *ss
>             }
>         }
>         None => default,
>     }
> }
> ```

No, actually if notice in search engines they fall to level 2 this is because in search engines the maximum `safe search level`  tends to be 2 and they are mean't to remove NSFW results but you see the sad truth is they most of the time don't remove NSFW results properly so to completely eliminate those results we have implemented a filter list method which allows the user to define custom filters (which is regex based) to remove explicit NSFW completely which gives more control to the user of what results they want to see/provide. I hope you got my point :slightly_smiling_face: .

src/models/server_models.rs
neon-mmd: > Yeah, I was getting annoying borrowed value doesn't live long enough when building a Cookie from other borrowed values.

For this I would suggest you need to set the cookie type to `Cookie<'_>` in the `search` function and this should solve the problem. 

Also, for more information on lifetimes in rust I would suggest watching this tutorials:

- https://piped.video/watch?v=juIINGuZyBc
- https://piped.video/watch?v=rAl-9HwD858

src/server/routes/search.rs
greengeckoz: I see, so they would need to modify the config file and set up their filter list as wanted to do that. Is there somewhere else in the code that enforces the safe search level to be that high?

src/models/server_models.rs
greengeckoz: I switched the Cookie back to using &str instead. I wasn't aware I could use a ref keyword in the match Some value to keep the reference in the correct scope.

src/server/routes/search.rs
```diff
@@ -48,17 +48,41 @@ pub async fn search(
...
+                Some(ref cookie_value) => {
+                    match serde_json::from_str(cookie_value.value()) {
+                        Ok(cookie) => cookie,
+                        // If there's an issue parsing the cookie's value, default to the config
+                        Err(_) => build_cookie(),
+                    }
+                }
+                // If there's no cookie saved, use the server's config
+                None => build_cookie(),
+            };
```
ibnz36: I believe you can use something like this here:
```rs
cookie
    .and_then(|c| serde_json::from_str(c.value()).ok())
    .unwrap_or_else(build_cookie);
```

ibnz36 (changes requested):

src/server/routes/search.rs
neon-mmd: > I believe you can use something like this here:
> 
> ```rust
> cookie
>     .and_then(|c| serde_json::from_str(c.value()).ok())
>     .unwrap_or_else(build_cookie);
> ```

Well, also, I think you will not need to write an extra closure for this you will just need to add the code from the `build_cookie` closure into the `unwrap_or_else`. What do you think?

ibnz36: You first need to convert the cookie to a serde something, don't you? Also unwrap_or_else here isn't taking a closure

neon-mmd: > You first need to convert the cookie to a serge something, don't you? Also unwrap_or_else here isn't taking a closure

Yes, you are right, but it you notice it carefully you will see that `build_cookie` in fact is a closure. :slightly_smiling_face: 

https://github.com/ddotthomas/websurfx/blob/a47e28587cd5e9e9bb93ce6520cb7ef35c436ae4/src/server/routes/search.rs#L52-L63

Also, I think the previous code had repetition, so we needed the `build_cookie` closure. As you can notice from the code snippet below.

https://github.com/ddotthomas/websurfx/blob/a47e28587cd5e9e9bb93ce6520cb7ef35c436ae4/src/server/routes/search.rs#L67-L78

 Now I think we don't need it anymore because there is only one instance `build_cookie` closure is being called. 
 
 ```rust
 cookie
    .and_then(|c| serde_json::from_str(c.value()).ok())
    .unwrap_or_else(build_cookie);
 ```
 
 So that's why I was suggesting to have change something like this:

```rust
cookie
    .and_then(|c| serde_json::from_str(c.value()).ok())
    .unwrap_or_else(
        server_models::Cookie::build(
            &config.style,
            config
                .upstream_search_engines
                .clone()
                .into_iter()
                .filter_map(|engine_map| engine_map.1.then_some(engine_map.0))
                .collect(),
            config.safe_search,
        )
    );
```

I hope you got my point :slightly_smiling_face:  .

ibnz36: Oh, I didn't see that. Yes, that's up to your code style. Sorry.

ibnz36: Also the code from the `build_cookie` seems a bit weird. Is that cloning and iteration needed for something specific?

greengeckoz: Yes, to solve the issue, the point of the pull request in the first place. We sort the engines list to make sure the list of engines matches between different users.

neon-mmd: > I believe you can use something like this here:
> 
> ```rust
> cookie
>     .and_then(|c| serde_json::from_str(c.value()).ok())
>     .unwrap_or_else(build_cookie);
> ```

Ok, after testing this implementation, I find that it doesn't work as it complains about `data borrowed by the current function`. So after playing around for a while I found a solution and here it is, and I would recommend using this instead:

In the **server_models.rs** file: 

```rust
/// A named struct which is used to deserialize the cookies fetched from the client side.
#[allow(dead_code)]
#[derive(Deserialize)]
pub struct Cookie<'a> {
    /// It stores the theme name used in the website.
    pub theme: Cow<'a, str>,
    /// It stores the colorscheme name used for the website theme.
    pub colorscheme: Cow<'a, str>,
    /// It stores the user selected upstream search engines selected from the UI.
    pub engines: Cow<'a, Vec<Cow<'a, str>>>,
    /// It stores the user selected safe search level from the UI.
    pub safe_search_level: u8,
}

impl<'a> Cookie<'a> {
    /// ---Add Appropriate Documentation Here Following The Rust Convention---
    pub fn new(style: &'a Style, safe_search_level: u8, mut engines: Vec<Cow<'a, str>>) -> Self {
        // ---Add Appropriate Comment Here---
        engines.sort();

        Self {
            theme: Cow::Borrowed(&style.theme),
            colorscheme: Cow::Borrowed(&style.colorscheme),
            engines: Cow::Owned(engines),
            safe_search_level,
        }
    }
}
```

In the **search.rs** file:

```rust
let cookie_value_or_config_values: server_models::Cookie<'_> = cookie_value
                .and_then(|cookie| serde_json::from_str(cookie.value()).ok())
                .unwrap_or_else(|| {
                    server_models::Cookie::new(
                        &config.style,
                        config.safe_search,
                        config
                            .upstream_search_engines
                            .iter()
                            .filter_map(|(key, value)| {
                                value.then_some(std::borrow::Cow::Borrowed(key.as_str()))
                            })
                            .collect(),
                    )
                });
```

For more information on `Copy on Write (Cow)` data type. I would suggest going through this standard library docs:

https://doc.rust-lang.org/std/borrow/enum.Cow.html 

**Note:** In the above implementation, I have also removed the `.clone()` and replaced it with the `into_iter()` for the `search.rs`. Also, in the `server_models.rs` cookie struct I have now replaced the previous data types with Cow data type as it seems to work best with the serialize and deserialize traits. Additionally, they also tend to be more memory efficient.

neon-mmd: > Also the code from the `build_cookie` seems a bit weird. Is that cloning and iteration needed for something specific?

Actually, I think the `cloning` part could be removed as I suggested from the above code suggestion. While for the iteration part it is required because in the config the value is stored as a key value pair containing the engine with the option true or false. So when we parse the config option it gets parsed as a `hashmap` which is used to then filter out search engines which are not selected by the user. 

Also, you may ask then `why we need the whole hashmap for this? we can just parse it and have it globally available in the Config struct, right?` While the reasoning behind using the `hashmap` is that in the `src/templates/partials/engines_tab`, We use this `hashmap` to display the user the saved config or cookie setting to show that their settings are being applied. Which is essential for good user experience. I hope you got my point :slightly_smiling_face:

src/server/routes/search.rs
neon-mmd: > I see, so they would need to modify the config file and set up their filter list as wanted to do that. Is there somewhere else in the code that enforces the safe search level to be that high?

Sorry for the late reply.

Yes, there are checks in the `aggregator.rs` file as well where it checks for the safe search level is equal to 3 or not and filters the results if the condition is true otherwise returns the search results as it is. :slightly_smiling_face:

src/server/routes/search.rs
```diff
@@ -237,23 +226,21 @@ fn is_match_from_filter_list(
...
-            }
+/// * `url_level` - Safe search level from the url.
+/// * `safe_search` - User's cookie, or the safe search level set by the server
+/// * `config_level` - Safe search level to fall back to
+fn get_safesearch_level(safe_search: &mut u8, url_level: &Option<u8>, config_level: u8) {
+    if let Some(search_level) = url_level {
+        if *search_level >= 3 {
+            *safe_search = config_level
+        } else {
+            *safe_search = *search_level;
```
neon-mmd: I would also suggest adding a fall back to cookie safe search level as well because you see there is a scenario where the engine will be configured by the user (server admin) using the config. Where the user might configure it with the safe search level of something other than to 3 or 4. Maybe because the user might not want to have a strictly SFW only server. In that case, if a user (user of the server) were to use such a server. And if let's say they don't like the default safe search level provided by the server admin of 2 or 0 and let's say they prefer 1. Then the user will have to each time set the safe search level from the search URL parameter, which you know is not great. Now if the user were to set it using the settings page which saves the safe search level as a cookie, this allows the user to set the safe search level only once. Hence, they will not need to change it again and again via the search URL parameter like they can still change it but, they might do it less often and prefer the more convenient settings page option. 

So, my suggestion would be to have a fallback in the else section. Where you can have a check for the none case for `search_level` variable, so when it is none it will fall back to using the cookie value if the user has set it. Thus overriding the value set by the server admin using the config. Otherwise, if it has not been set by the user then you know obviously you will fall back to the config value because the value set by the server admin was not overridden. So we will use it then. Also, if the `search_level` was provided using the search parameter, then we pass it as it is because it will override both the cookie and the value set by the server admin. I hope you got my idea :slightly_smiling_face: .

neon-mmd (changes requested):

neon-mmd (approved):
Thanks :heart: again for working on this PR. We really appeciate it :+1: Since everything looka good. We will merge the PR as soon as possible :slightly_smiling_face: .

alamin655 (approved):

PR #473, opened 2023-12-31
✨ `Bing` for the search engine

### LINKED ISSUE #316: ✨ `Bing` for the search engine
Provide the `bing` search engine for the `websurfx` search engine.  To start working on this issue, you will need to create a new file with the name of the desired engine in the folder `src/engines/` which is located under the codebase (`websurfx` directory). Then in the file, just create a new struct with the name of the engine and then create two `impl` blocks for the struct. One for implementing the `SearchEngine` trait on it and the other for implementing the `parsers` to be used for aggregating results.

> **Note**

jfvillablanca:
## What does this PR do?

Adds `Bing` search engine to websurfx

Expand the search results horizon with `Bing` which helps improve user experience.

Enable `Bing` in config.lua or via the settings UI.
## Related issues
Closes #316

### DIFF
--- a/src/engines/bing.rs
+++ b/src/engines/bing.rs
@@ -1,0 +1,124 @@
+//! The `bing` module handles the scraping of results from the bing search engine
+//! by querying the upstream bing search engine with user provided query and with a page
+//! number if provided.
+
+use std::collections::HashMap;
+
+use regex::Regex;
+use reqwest::header::HeaderMap;
+use reqwest::Client;
+use scraper::Html;
+
+use crate::models::aggregation_models::SearchResult;
+
+use crate::models::engine_models::{EngineError, SearchEngine};
+
+use error_stack::{Report, Result, ResultExt};
+
+use super::search_result_parser::SearchResultParser;
+
+/// A new Bing engine type defined in-order to implement the `SearchEngine` trait which allows to
+/// reduce code duplication as well as allows to create vector of different search engines easily.
+pub struct Bing {
+    /// The parser, used to interpret the search result.
+    parser: SearchResultParser,
+}
+
+impl Bing {
+    /// Creates the Bing parser.
+    pub fn new() -> Result<Self, EngineError> {
+        Ok(Self {
+            parser: SearchResultParser::new(
+                ".b_results",
+                ".b_algo",
+                "h2 a",
+                ".tpcn a.tilk",
+                ".b_caption p",
+            )?,
+        })
+    }
+}
+
+#[async_trait::async_trait]
+impl SearchEngine for Bing {
+    async fn results(
+        &self,
+        query: &str,
+        page: u32,
+        user_agent: &str,
+        client: &Client,
+        _safe_search: u8,
+    ) -> Result<HashMap<String, SearchResult>, EngineError> {
+        // Bing uses `start results from this number` convention
+        // So, for 10 results per page, page 0 starts at 1, page 1
+        // starts at 11, and so on.
+        let results_per_page = 10;
+        let start_result = results_per_page * page + 1;
+
+        let url: String = match page {
+            0 => {
+                format!("https://www.bing.com/search?q={query}")
+            }
+            _ => {
+                format!("https://www.bing.com/search?q={query}&first={start_result}")
+            }
+        };
+
+        let query_params: Vec<(&str, &str)> = vec![
+            ("_EDGE_V", "1"),
+            ("SRCHD=AF", "NOFORM"),
+            ("_Rwho=u", "d"),
+            ("bngps=s", "0"),
+            ("_UR=QS=0&TQS", "0"),
+            ("_UR=QS=0&TQS", "0"),
+        ];
+
+        let mut cookie_string = String::new();
+        for (k, v) in &query_params {
+            cookie_string.push_str(&format!("{k}={v}; "));
+        }
+
+        let header_map = HeaderMap::try_from(&HashMap::from([
+            ("USER_AGENT".to_string(), user_agent.to_string()),
+            ("REFERER".to_string(), "https://google.com/".to_string()),
+            (
+                "CONTENT_TYPE".to_string(),
+                "application/x-www-form-urlencoded".to_string(),
+            ),
+            ("COOKIE".to_string(), cookie_string),
+        ]))
+        .change_context(EngineError::UnexpectedError)?;
+
+        let document: Html = Html::parse_document(
+            &Bing::fetch_html_from_upstream(self, &url, header_map, client).await?,
+        );
+
+        // Bing is very aggressive in finding matches
+        // even with the most absurd of queries. ".b_algo" is the
+        // class for the list item of results
+        if let Some(no_result_msg) = self.parser.parse_for_no_results(&document).nth(0) {
+            if no_result_msg
+                .value()
+                .attr("class")
+                .map(|classes| classes.contains("b_algo"))
+                .unwrap_or(false)
+            {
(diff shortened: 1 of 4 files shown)


src/engines/bing.rs
```diff
@@ -0,0 +1,109 @@
...
+
+        let header_map = HeaderMap::try_from(&HashMap::from([
+            ("USER_AGENT".to_string(), user_agent.to_string()),
+            ("REFERER".to_string(), "https://google.com/".to_string()),
+            (
+                "CONTENT_TYPE".to_string(),
+                "application/x-www-form-urlencoded".to_string(),
+            ),
+        ]))
+        .change_context(EngineError::UnexpectedError)?;
```
neon-mmd: I think I would suggest adding a `cookie` header because this helps prevent being detected as a bot by the upstream search engine. 

Also, the cookie for bing will look something like the following:
```
MUID=2D0BD206B856600B0802C1DBB95F61B9; _EDGE_V=1; MUIDB=2D0BD206B856600B0802C1DBB95F61B9; SRCHD=AF=NOFORM; SRCHUID=V=2&GUID=B68AA9D86D6F45C5A891D99D8902C941&dmnchg=1; _HPVN=CS=eyJQbiI6eyJDbiI6MiwiU3QiOjAsIlFzIjowLCJQcm9kIjoiUCJ9LCJTYyI6eyJDbiI6MiwiU3QiOjAsIlFzIjowLCJQcm9kIjoiSCJ9LCJReiI6eyJDbiI6MiwiU3QiOjAsIlFzIjowLCJQcm9kIjoiVCJ9LCJBcCI6dHJ1ZSwiTXV0ZSI6dHJ1ZSwiTGFkIjoiMjAyMy0xMi0zMVQwMDowMDowMFoiLCJJb3RkIjowLCJHd2IiOjAsIlRucyI6MCwiRGZ0IjpudWxsLCJNdnMiOjAsIkZsdCI6MCwiSW1wIjo0LCJUb2JuIjowfQ==; _Rwho=u=d; ai_session=TB676+0gO13S3BW5AYVW3d|1704013243633|1704013243633; BCP=AD=0&AL=0&SM=0; ipv6=hit=1704016843140&t=4; _EDGE_S=SID=13B4DE4C147D60EB12DECDB415A661CB&mkt=en-kw; USRLOC=HS=1&ELOC=LAT=29.333396911621094|LON=47.99052810668945|N=Mirqab%2C%20Al%20Asimah|ELT=1|; SRCHUSR=DOB=20231231&T=1704013237000&TPC=1704013248000; bngps=s=0; _UR=QS=0&TQS=0; _RwBf=r=0&ilt=10&ihpd=1&ispd=6&rc=18&rb=0&gb=0&rg=200&pc=15&mtu=0&rbb=0&g=0&cid=&clo=0&v=10&l=2023-12-31T08:00:00.0000000Z&lft=0001-01-01T00:00:00.0000000&aof=0&ard=0001-01-01T00:00:00.0000000&o=2&p=&c=&t=0&s=0001-01-01T00:00:00.0000000+00:00&ts=2023-12-31T09:04:39.0642186+00:00&rwred=0&wls=&wlb=&wle=&ccp=&lka=0&lkt=0&aad=0&TH=; _SS=SID=13B4DE4C147D60EB12DECDB415A661CB&R=18&RB=0&GB=0&RG=200&RP=15; SRCHHPGUSR=SRCHLANG=en&IG=5AFEB9E3084249A9AF2DADF971C5965F&PV=6.6.8&BRW=NOTP&BRH=M&CW=819&CH=703&SCW=1164&SCH=3480&DPR=1.0&UTC=180&DM=1&HV=1704013480&WTS=63839610037&PRVCW=1325&PRVCH=703&NEWWND=0&NRSLT=10c&LSL=0&AS=1&NNT=0&HAP=2&VSRO=0&CHTRSP=3&WEBTHEME=1
```

src/engines/bing.rs
```diff
@@ -0,0 +1,109 @@
...
+            .parse_for_results(&document, |title, url, desc| {
+                Some(SearchResult::new(
+                    &re_strong.replace_all(title.inner_html().trim(), ""),
+                    url.value().attr("href").unwrap(),
+                    &re_strong.replace_all(&re_span.replace_all(desc.inner_html().trim(), ""), ""),
+                    &["bing"],
+                ))
+            })
+    }
+}
```
neon-mmd:
```suggestion

        // scrape all the results from the html
        self.parser
            .parse_for_results(&document, |title, url, desc| {
                Some(SearchResult::new(
                    &re_strong.replace_all(title.inner_html().trim(), ""),
                    url.value().attr("href").unwrap(),
                    desc.inner_html().trim(),
                    &["bing"],
                ))
            })
    }
}
```

I think you don't need to replace `span` and `strong` tags because they will be rendered on the search engine page. Also, having this would also be necessary as this will highlight what matched in the description of the site because of which these results were shown. This can add a little aesthetic style which can provide a good user experience to the user. So I would suggest leaving it :slightly_smiling_face: .

neon-mmd (changes requested):
Thanks :heart: for working on this PR. We really appreciate it :+1: but we would like to suggest a few changes before merge this PR. :slightly_smiling_face:

src/engines/bing.rs
jfvillablanca: I did not include the cookies because their cookies are very long and generated per-user. How about we include the shorter ones like `_EDGE_v=1`, `SRCHD=AF=NOFORM`, etc, what do you think?

src/engines/bing.rs
jfvillablanca: For the `strong` tags, I figured it was necessary to sanitize and to maintain uniformity among the search engines, i.e. other search engines do not insert semantic tags on their results. But if we want that in our results, I'll remove it

For the `span` tags, I believe that one is necessary to be removed because it removes a Bing-specific element not relevant to the results

### Screenshot from Bing. The _Web_ text is inside the span which I removed.



### Screenshot from Websurfx with the `strong` and `span` tags included

src/engines/bing.rs
neon-mmd: > I did not include the cookies because their cookies are very long and generated per-user. How about we include the shorter ones like `_EDGE_v=1`, `SRCHD=AF=NOFORM`, etc, what do you think?

Ok I see, I think then I would suggest adding them and checking if everything works still. Also, if there are more cookie parameters then I would suggest exploring them and adding them too. :slightly_smiling_face:

src/engines/bing.rs
neon-mmd: > For the `strong` tags, I figured it was necessary to sanitize and to maintain uniformity among the search engines, i.e. other search engines do not insert semantic tags on their results. But if we want that in our results, I'll remove it
> 
> For the `span` tags, I believe that one is necessary to be removed because it removes a Bing-specific element not relevant to the results
> 
> ### Screenshot from Bing. The _Web_ text is inside the span which I removed.
> 
> 
> ### Screenshot from Websurfx with the `strong` and `span` tags included
> 

Ok after seeing the picture I agree with you now that those `spans` are kind of useless and needs to be removed but for the `strong` tags I would suggest keeping them because I think as mentioned earlier when they get rendered on the search page they will add an aesthetic style like showing the user what in the description of the site matched their query which can help provide a better user experience. :slightly_smiling_face:

neon-mmd (approved):
Thanks :heart: again for working on this PR. We really appreciate it :+1: Since everything looks good. We will merge this PR right away :rocket: :slightly_smiling_face: .

neon-mmd:
@Mergifyio squash

PR #519, opened 2024-02-06
:sparkles: `Json` static configurarion file support 

### LINKED ISSUE #517: ✨ Support json for the config format
### Description

This makes it easier to configure websurfx with the Nix language.

Yes

### LINKED ISSUE #518: ✨ restructure the config to have different layers for better composability
### Description

Right now all of the config options are in one level (lua globals) this makes it more confusing to know which category each setting is related to. In addition, we can use mlua's serde feature to de-serialize directly from that structure. 

### Screenshots

_No response_

Yes

### Additional information

_No response_

Kranzes:
## What does this PR do?
 - Restructure config file to have more nested fields for better composability
 - Implement the Default trait for the Config struct to support falling back to the default configuration settings if they are not present in the config file or if a config file was not found (previously you had copy over the default config file during installation otherwise the program would just panic in a really ugly way)
 - Support both JSON and Lua for the config file format using Serde Deserialization

`cargo test` - Lua config
`cargo test --features json-config, memory-cache. --no-default-features` JSON config

## Additional Notes

WIP: still need to update the docs

Idea: Ideally we should not repeat ourselves when it comes to the documentation, at the moment we maintain the documentation for the configuration options across multiple files in the `docs` directory and they are often duplicated. We should look into having a way to generate nice looking documentation directly from the Config struct, or only maintain one instance of the `config.lua` file and link to it where we currently duplicate the text.

## Related issues
Closes #517
Closes #518


neon-mmd (changes requested):
Thanks :heart: for working on such a wonderful PR. We really appreciate it :+1: Keep it up!! But we would like to suggest a few changes before we merge this PR in. 

I think I would suggest not having default implementations because during installation the project will make sure to install the default config in the right place, so having `config file not found error` would be extremely rare. Moreover, I think panicking on the config file would also make sense because if the config file is not found then we should not technically run the server at all. The idea is similar too many programs that do the same thing as many tiling window managers, web browsers like qutebrowser, etc. So I think panicking on the config file not found might not be very problematic, instead it would be valid to fail on it. Additionally, not having default implementation for struct would also reduce the amount of code required in config parsing too. :slightly_smiling_face:

Kranzes:
> because during installation the project will make sure to install the default config in the right place

Running `cargo build/install` does not also install the config file, that makes your point invalid imo. I strongly believe you should never, ever panic on a user, at least print to STDERR and exit with a non-zero exit code rather than just a lazy use of `unwrap`. (which is considered a an anti-pattern by many people in the Rust community)

>so having config file not found error would be extremely rare.

I can tell you that it's not. This is exactly how I stumbled upon `websurfx` to begin with, someone reached out to me to help them why the program just crashes on them after they built it with Nix. They were used to programs having internal fallbacks if a config is missing. This is not the first time I encountered this as a package maintainer in the biggest cross-platform repository in the world. This 99% of the times comes from poorly written closed-source software that on-purpose want you to reach out to them for paid official support.

> we should not technically run the server at all

Again, you're making too many assumptions on the setup of the user, that's bad design. If you want your program to be successful and used more often, don't be this intrusive towards the user's setup. The situation is even worse if you can't even specify on the command line the path to the `config.lua`. You shouldn't assume the user has the exact filesystem structure as yours.

> Not having default implementation for struct would also reduce the amount of code required in config parsing too.

That's not true, at least not with Serde. With this PR the implementation the parsing of the config is legit smaller than before.

> many programs that do the same thing as many tiling window managers, web browsers like qutebrowser,

Yeah, and they are poorly designed, look at how common it is for these programs to get issues from beginners about "why is there only a blank screen", "why does it just crash". Your program should be robust and have recoverable errors/fallbacks, If that's something you don't understand then, no offense, read a book.

---

As previously mentioned in the Discord, I will most likely not use `websurfx` for other reasons, I created this PR to help out and I do not have the will to implement the documentation part for it, so I can either close this PR or let it get stale because I don't think it's worth it for me to deal with more impedance on a project I don't even care enough to personally use.

websurfx/config.lua
```diff
@@ -1,69 +1,41 @@
...
-    LibreX = false,
-    Mojeek = false,
-    Bing = false,
-} -- select the upstream search engines from which the results should be fetched.
+    style = {
+   	colorscheme = "catppuccin-mocha",
+   	theme = "simple",
+    	animation = "simple-frosted-glow"
+    }
+}
```
neon-mmd, line 41: I would suggest providing documentation in the config file as well. This will allow the user to view the different config option documentation within the config file itself, and the user will not need to look at the docs again and again by visiting the project. Also, this will also allow the user to view the docs offline too without them being actively connected to the internet. :slightly_smiling_face:

neon-mmd (changes requested):

neon-mmd:
> > because during installation the project will make sure to install the default config in the right place
> 
> Running `cargo build/install` does not also install the config file, that makes your point invalid imo. I strongly believe you should never, ever panic on a user, at least print to STDERR and exit with a non-zero exit code rather than just a lazy use of `unwrap`. (which is considered a an anti-pattern by many people in the Rust community)
> 
> > so having config file not found error would be extremely rare.
> 
> I can tell you that it's not. This is exactly how I stumbled upon `websurfx` to begin with, someone reached out to me to help them why the program just crashes on them after they built it with Nix. They were used to programs having internal fallbacks if a config is missing. This is not the first time I encountered this as a package maintainer in the biggest cross-platform repository in the world. This 99% of the times comes from poorly written closed-source software that on-purpose want you to reach out to them for paid official support.
> 
> > we should not technically run the server at all
> 
> Again, you're making too many assumptions on the setup of the user, that's bad design. If you want your program to be successful and used more often, don't be this intrusive towards the user's setup. The situation is even worse if you can't even specify on the command line the path to the `config.lua`. You shouldn't assume the user has the exact filesystem structure as yours.
> 
> > Not having default implementation for struct would also reduce the amount of code required in config parsing too.
> 
> That's not true, at least not with Serde. With this PR the implementation the parsing of the config is legit smaller than before.
> 
> > many programs that do the same thing as many tiling window managers, web browsers like qutebrowser,
> 
> Yeah, and they are poorly designed, look at how common it is for these programs to get issues from beginners about "why is there only a blank screen", "why does it just crash". Your program should be robust and have recoverable errors/fallbacks, If that's something you don't understand then, no offense, read a book.
> 
> As previously mentioned in the Discord, I will most likely not use `websurfx` for other reasons, I created this PR to help out and I do not have the will to implement the documentation part for it, so I can either close this PR or let it get stale because I don't think it's worth it for me to deal with more impedance on a project I don't even care enough to personally use.

Ok I see cool thanks :heart: for taking the time to explain. Now after reanalyzing things as you had suggested I really see your point. So I agree lets have this changes then. :slightly_smiling_face:

Kranzes:
Still have to figure out the documentation aspect, but I think I should also make it so if no config is found the program will write one directly, so basically serialize from the struct to a lua file, rather than it be invisible to the users.

neon-mmd:
> Still have to figure out the documentation aspect, but I think I should also make it so if no config is found the program will write one directly, so basically serialize from the struct to a lua file, rather than it be invisible to the users.

Well, we are already falling back to default values for the options when no config is found. So I wonder why would we need to generate one? Or wait I have another idea to maybe instead of falling back to defaults why don't we auto generate a config when no config is found with the default values instead of panicking and use the newly generated config instead? What do you think? :slightly_smiling_face:

Kranzes:
> Well, we are already falling back to default values for the options when no config is found. So I wonder why would we need to generate one? Or wait I have another idea to maybe instead of falling back to defaults why don't we auto generate a config when no config is found with the default values instead of panicking and use the newly generated config instead? What do you think? 🙂

That's exactly what I am trying to do, I managed to do it for JSON really easily, not so much for Lua though, but I will keep you updated.

Kranzes:
I am closing this out of loss of motivation due to odd design choices made to the code-base initially that make all of this a big burden. e.g. lua + manual struct initialization + not using a smart configuration library + bad error handling + DIY directories library and more

I was able to get a TOML config file generated with the comments from the docstrings using confique (I sent in the discord a snippet).

bye bye :wave:

PR #529, opened 2024-02-19
:sparkles: Config option to configure the https adaptive window size for requests

### LINKED ISSUE #527: ✨ Config option to use adaptive `http` window size for requests
#### Work Expected From The Issue
Provide a config option to allow user to enable `http` adaptive window size for requests for fetching search results from upstream search engines.

The issue expects the following files to be changed:
- `src/results/aggregator.rs`
- `src/parser/config.rs`
- `websurfx/config.lua`

> [!Note]
> All the files that are expected to be changed are located under the codebase (`websurfx` directory).

#### Reason Behind These Changes

The reason behind having these changes is to allow the user to enable `http` adaptive window sizes option for requests which improves the performance of requests on reliable/stable internet connections which in turn improves user experience ny reducing latency for responses.

#### Sample Code
The sample codes for two of the files from the above mentioned files has been provided below:

**aggregator.rs**
```diff
//! This module provides the functionality to scrape and gathers all the results from the upstream
//! search engines and then removes duplicate results.

use super::user_agent::random_user_agent;
use crate::handler::{file_path, FileType};
use crate::models::{
    aggregation_models::{EngineErrorInfo, SearchResult, SearchResults},
    engine_models::{EngineError, EngineHandler},
};
use error_stack::Report;
use futures::stream::FuturesUnordered;
use regex::Regex;
use reqwest::{Client, ClientBuilder};
use std::time::{SystemTime, UNIX_EPOCH};
use std::{fs::File, io::BufRead};
use std::{
    io::{BufReader, Read},
    time::Duration,
};
use tokio::task::JoinHandle;

/// A constant for holding the prebuilt Client globally in the app.
static CLIENT: std::sync::OnceLock<Client> = std::sync::OnceLock::new();

/// Aliases for long type annotations

type FutureVec =
    FuturesUnordered<JoinHandle<Result<Vec<(String, SearchResult)>, Report<EngineError>>>>;

/// The function aggregates the scraped results from the user-selected upstream search engines.
/// These engines can be chosen either from the user interface (UI) or from the configuration file.
/// The code handles this process by matching the selected search engines and adding them to a vector.
/// This vector is then used to create an asynchronous task vector using `tokio::spawn`, which returns
/// a future. This future is awaited in another loop. Once the results are collected, they are filtered
/// to remove any errors and ensure only proper results are included. If an error is encountered, it is
/// sent to the UI along with the name of the engine and the type of error. This information is finally
/// placed in the returned `SearchResults` struct.
///
/// Additionally, the function eliminates duplicate results. If two results are identified as coming from
/// multiple engines, their names are combined to indicate that the results were fetched from these upstream
/// engines. After this, all the data in the `Vec` is removed and placed into a struct that contains all
/// the aggregated results in a vector. Furthermore, the query used is also added to the struct. This step is
/// necessary to ensure that the search bar in the search remains populated even when searched from the query URL.
///
/// Overall, this function serves to aggregate scraped results from user-selected search engines, handling errors,
/// removing duplicates, and organizing the data for display in the UI.
///
/// # Example:
///
/// If you search from the url like `https://127.0.0.1/search?q=huston` then the search bar should
/// contain the word huston and not remain empty.
///
/// # Arguments
///
/// * `query` - Accepts a string to query with the above upstream search engines.
/// * `page` - Accepts an u32 page number.
/// * `random_delay` - Accepts a boolean value to add a random delay before making the request.
/// * `debug` - Accepts a boolean value to enable or disable debug mode option.
/// * `upstream_search_engines` - Accepts a vector of search engine names which was selected by the
/// * `request_timeout` - Accepts a time (secs) as a value which controls the server request timeout.
/// user through the UI or the config file.
///
/// # Error
///
/// Returns an error a reqwest and scraping selector errors if any error occurs in the results
/// function in either `searx` or `duckduckgo` or both otherwise returns a `SearchResults struct`
/// containing appropriate values.
pub async fn aggregate(
    query: &str,
    page: u32,
    random_delay: bool,
    debug: bool,
    upstream_search_engines: &[EngineHandler],
    request_timeout: u8,
    safe_search: u8,
+  https_adaptive_window_size: bool,
) -> Result<SearchResults, Box<dyn std::error::Error>> {
    let client = CLIENT.get_or_init(|| {
        ClientBuilder::new()
            .timeout(Duration::from_secs(request_timeout as u64)) // Add timeout to request to avoid DDOSing the server
+          .http2_adaptive_window(https_adaptive_window_size)
            .https_only(true)
            .gzip(true)
            .brotli(true)
            .build()
            .unwrap()
    });

    let user_agent: &str = random_user_agent();

    // Add a random delay before making the request.
    if random_delay || !debug {
        let nanos = SystemTime::now().duration_since(UNIX_EPOCH)?.subsec_nanos() as f32;
        let delay = ((nanos / 1_0000_0000 as f32).floor() as u64) + 1;
        tokio::time::sleep(Duration::from_secs(delay)).await;
    }

    let mut names: Vec<&str> = Vec::with_capacity(0);

    // create tasks for upstream result fetching
    let tasks: FutureVec = FutureVec::new();

    for engine_handler in upstream_search_engines {
        let (name, search_engine) = engine_handler.to_owned().into_name_engine();
        names.push(name);
        let query: String = query.to_owned();
        tasks.push(tokio::spawn(async move {
            search_engine
                .results(&query, page, user_agent, client, safe_search)
                .await
        }));
    }

    // get upstream responses
    let mut responses = Vec::with_capacity(tasks.len());

    for task in tasks {
        if let Ok(result) = task.await {
            responses.push(result)
        }
    }

    // aggregate search results, removing duplicates and handling errors the upstream engines returned
    let mut result_map: Vec<(String, SearchResult)> = Vec::new();
    let mut engine_errors_info: Vec<EngineErrorInfo> = Vec::new();

    let mut handle_error = |error: &Report<EngineError>, engine_name: &'static str| {
        log::error!("Engine Error: {:?}", error);
        engine_errors_info.push(EngineErrorInfo::new(
            error.downcast_ref::<EngineError>().unwrap(),
            engine_name,
        ));
    };

    for _ in 0..responses.len() {
        let response = responses.pop().unwrap();
        let engine = names.pop().unwrap();

        if result_map.is_empty() {
            match response {
                Ok(results) => result_map = results,
                Err(error) => handle_error(&error, engine),
            };
            continue;
        }

        match response {
            Ok(result) => {
                result.into_iter().for_each(|(key, value)| {
                    match result_map.iter().find(|(key_s, _)| key_s == &key) {
                        Some(value) => value.1.to_owned().add_engines(engine),
                        None => result_map.push((key, value)),
                    };
                });
            }
            Err(error) => handle_error(&error, engine),
        };
    }

    if safe_search >= 3 {
        let mut blacklist_map: Vec<(String, SearchResult)> = Vec::new();
        filter_with_lists(
            &mut result_map,
            &mut blacklist_map,
            file_path(FileType::BlockList)?,
        )?;

        filter_with_lists(
            &mut blacklist_map,
            &mut result_map,
            file_path(FileType::AllowList)?,
        )?;

        drop(blacklist_map);
    }

    let results: Vec<SearchResult> = result_map.iter().map(|(_, value)| value.clone()).collect();

    Ok(SearchResults::new(results, &engine_errors_info))
}

/// Filters a map of search results using a list of regex patterns.
///
/// # Arguments
///
/// * `map_to_be_filtered` - A mutable reference to a `Vec` of search results to filter, where the filtered results will be removed from.
/// * `resultant_map` - A mutable reference to a `Vec` to hold the filtered results.
/// * `file_path` - A `&str` representing the path to a file containing regex patterns to use for filtering.
///
/// # Errors
///
/// Returns an error if the file at `file_path` cannot be opened or read, or if a regex pattern is invalid.
pub fn filter_with_lists(
    map_to_be_filtered: &mut Vec<(String, SearchResult)>,
    resultant_map: &mut Vec<(String, SearchResult)>,
    file_path: &str,
) -> Result<(), Box<dyn std::error::Error>> {
    let mut reader = BufReader::new(File::open(file_path)?);

    for line in reader.by_ref().lines() {
        let re = Regex::new(line?.trim())?;

        // Iterate over each search result in the map and check if it matches the regex pattern
        for (index, (url, search_result)) in map_to_be_filtered.clone().into_iter().enumerate() {
            if re.is_match(&url.to_lowercase())
                || re.is_match(&search_result.title.to_lowercase())
                || re.is_match(&search_result.description.to_lowercase())
            {
                // If the search result matches the regex pattern, move it from the original map to the resultant map
                resultant_map.push(map_to_be_filtered.remove(index));
            }
        }
    }

    Ok(())
}

#[cfg(test)]
mod tests {
    use super::*;
    use smallvec::smallvec;
    use std::io::Write;
    use tempfile::NamedTempFile;

    #[test]
    fn test_filter_with_lists() -> Result<(), Box<dyn std::error::Error>> {
        // Create a map of search results to filter
        let mut map_to_be_filtered = Vec::new();
        map_to_be_filtered.push((
            "https://www.example.com".to_owned(),
            SearchResult {
                title: "Example Domain".to_owned(),
                url: "https://www.example.com".to_owned(),
                description: "This domain is for use in illustrative examples in documents."
                    .to_owned(),
                engine: smallvec!["Google".to_owned(), "Bing".to_owned()],
            },
        ));
        map_to_be_filtered.push((
            "https://www.rust-lang.org/".to_owned(),
            SearchResult {
                title: "Rust Programming Language".to_owned(),
                url: "https://www.rust-lang.org/".to_owned(),
                description: "A systems programming language that runs blazingly fast, prevents segfaults, and guarantees thread safety.".to_owned(),
                engine: smallvec!["Google".to_owned(), "DuckDuckGo".to_owned()],
            },)
        );

        // Create a temporary file with regex patterns
        let mut file = NamedTempFile::new()?;
        writeln!(file, "example")?;
        writeln!(file, "rust")?;
        file.flush()?;

        let mut resultant_map = Vec::new();
        filter_with_lists(
            &mut map_to_be_filtered,
            &mut resultant_map,
            file.path().to_str().unwrap(),
        )?;

        assert_eq!(resultant_map.len(), 2);
        assert!(resultant_map
            .iter()
            .any(|(key, _)| key == "https://www.example.com"));
        assert!(resultant_map
            .iter()
            .any(|(key, _)| key == "https://www.rust-lang.org/"));
        assert_eq!(map_to_be_filtered.len(), 0);

        Ok(())
    }

    #[test]
    fn test_filter_with_lists_wildcard() -> Result<(), Box<dyn std::error::Error>> {
        let mut map_to_be_filtered = Vec::new();
        map_to_be_filtered.push((
            "https://www.example.com".to_owned(),
            SearchResult {
                title: "Example Domain".to_owned(),
                url: "https://www.example.com".to_owned(),
                description: "This domain is for use in illustrative examples in documents."
                    .to_owned(),
                engine: smallvec!["Google".to_owned(), "Bing".to_owned()],
            },
        ));
        map_to_be_filtered.push((
            "https://www.rust-lang.org/".to_owned(),
            SearchResult {
                title: "Rust Programming Language".to_owned(),
                url: "https://www.rust-lang.org/".to_owned(),
                description: "A systems programming language that runs blazingly fast, prevents segfaults, and guarantees thread safety.".to_owned(),
                engine: smallvec!["Google".to_owned(), "DuckDuckGo".to_owned()],
            },
        ));

        // Create a temporary file with a regex pattern containing a wildcard
        let mut file = NamedTempFile::new()?;
        writeln!(file, "ex.*le")?;
        file.flush()?;

        let mut resultant_map = Vec::new();

        filter_with_lists(
            &mut map_to_be_filtered,
            &mut resultant_map,
            file.path().to_str().unwrap(),
        )?;

        assert_eq!(resultant_map.len(), 1);
        assert!(resultant_map
            .iter()
            .any(|(key, _)| key == "https://www.example.com"));
        assert_eq!(map_to_be_filtered.len(), 1);
        assert!(map_to_be_filtered
            .iter()
            .any(|(key, _)| key == "https://www.rust-lang.org/"));

        Ok(())
    }

    #[test]
    fn test_filter_with_lists_file_not_found() {
        let mut map_to_be_filtered = Vec::new();

        let mut resultant_map = Vec::new();

        // Call the `filter_with_lists` function with a non-existent file path
        let result = filter_with_lists(
            &mut map_to_be_filtered,
            &mut resultant_map,
            "non-existent-file.txt",
        );

        assert!(result.is_err());
    }

    #[test]
    fn test_filter_with_lists_invalid_regex() {
        let mut map_to_be_filtered = Vec::new();
        map_to_be_filtered.push((
            "https://www.example.com".to_owned(),
            SearchResult {
                title: "Example Domain".to_owned(),
                url: "https://www.example.com".to_owned(),
                description: "This domain is for use in illustrative examples in documents."
                    .to_owned(),
                engine: smallvec!["Google".to_owned(), "Bing".to_owned()],
            },
        ));

        let mut resultant_map = Vec::new();

        // Create a temporary file with an invalid regex pattern
        let mut file = NamedTempFile::new().unwrap();
        writeln!(file, "example(").unwrap();
        file.flush().unwrap();

        let result = filter_with_lists(
            &mut map_to_be_filtered,
            &mut resultant_map,
            file.path().to_str().unwrap(),
        );

        assert!(result.is_err());
    }
}
```

**config.lua**
```diff
 -- ### General ###
logging = true -- an option to enable or disable logs.
debug = false -- an option to enable or disable debug mode.
threads = 10 -- the amount of threads that the app will use to run (the value should be greater than 0).

 -- ### Server ###
port = "8080" -- port on which server should be launched
binding_ip = "127.0.0.1" --ip address on the which server should be launched.
production_use = false -- whether to use production mode or not (in other words this option should be used if it is to be used to host it on the server to provide a service to a large number of users (more than one))
 -- if production_use is set to true
 -- There will be a random delay before sending the request to the search engines, this is to prevent DDoSing the upstream search engines from a large number of simultaneous requests.
request_timeout = 30 -- timeout for the search requests sent to the upstream search engines to be fetched (value in seconds).
+https_adaptive_window_size = false -- whether to enable or disable https adaptive window sizing to reduce the response latency between for each request between the server and ths client (this option should only be used if you have a reliable/stable internet connection).
rate_limiter = {
	number_of_requests = 20, -- The number of request that are allowed within a provided time limit.
	time_limit = 3, -- The time limit in which the quantity of requests that should be accepted.
}

 -- ### Search ###
 -- Filter results based on different levels. The levels provided are:
 -- {{
 -- 0 - None
 -- 1 - Low
 -- 2 - Moderate
 -- 3 - High
 -- 4 - Aggressive
 -- }}
safe_search = 2

 -- ### Website ###
 -- The different colorschemes provided are:
 -- {{
 -- catppuccin-mocha
 -- dark-chocolate
 -- dracula
 -- gruvbox-dark
 -- monokai
 -- nord
 -- oceanic-next
 -- one-dark
 -- solarized-dark
 -- solarized-light
 -- tokyo-night
 -- tomorrow-night
 -- }}
colorscheme = "catppuccin-mocha" -- the colorscheme name which should be used for the website theme
 -- The different themes provided are:
 -- {{
 -- simple
 -- }}
theme = "simple" -- the theme name which should be used for the website
 -- The different animations provided are:
 -- {{
 -- simple-frosted-glow
 -- }}
animation = "simple-frosted-glow" -- the animation name which should be used with the theme or `nil` if you don't want any animations.

 -- ### Caching ###
redis_url = "redis://127.0.0.1:8082" -- redis connection url address on which the client should connect on.
cache_expiry_time = 600 -- This option takes the expiry time of the search results (value in seconds and the value should be greater than or equal to 60 seconds).
 -- ### Search Engines ###
upstream_search_engines = {
    DuckDuckGo = true,
    Searx = false,
    Brave = false,
    Startpage = false,
    LibreX = false,
    Mojeek = false,
    Bing = false,
} -- select the upstream search engines from which the results should be fetched.
```

greengeckoz:
## What does this PR do?

Provided a new config option to allow user enable/disable the `https` window size setting for the requests fetching the search results from the upstream search engines.

These change is essential as it allows the user to enable `https` adaptive window sizes option for requests which improves the performance of requests on reliable/stable internet connections which in turn improves user experience ny reducing latency for responses.

Check TCP SYN and ACK packets in wireshark for TCP Window Size scaling

## Related issues

Closes #527

### DIFF
--- a/src/results/aggregator.rs
+++ b/src/results/aggregator.rs
@@ -2,6 +2,7 @@
 //! search engines and then removes duplicate results.
 
 use super::user_agent::random_user_agent;
+use crate::config::parser::Config;
 use crate::handler::{file_path, FileType};
 use crate::models::{
     aggregation_models::{EngineErrorInfo, SearchResult, SearchResults},
@@ -66,20 +67,17 @@ type FutureVec = Vec<JoinHandle<Result<HashMap<String, SearchResult>, Report<Eng
 pub async fn aggregate(
     query: &str,
     page: u32,
-    random_delay: bool,
-    debug: bool,
+    config: &Config,
     upstream_search_engines: &[EngineHandler],
-    request_timeout: u8,
     safe_search: u8,
-    adaptive_window: bool,
 ) -> Result<SearchResults, Box<dyn std::error::Error>> {
     let client = CLIENT.get_or_init(|| {
         ClientBuilder::new()
-            .timeout(Duration::from_secs(request_timeout as u64)) // Add timeout to request to avoid DDOSing the server
+            .timeout(Duration::from_secs(config.request_timeout as u64)) // Add timeout to request to avoid DDOSing the server
             .https_only(true)
             .gzip(true)
             .brotli(true)
-            .http2_adaptive_window(adaptive_window)
+            .http2_adaptive_window(config.adaptive_window)
             .build()
             .unwrap()
     });
@@ -87,7 +85,7 @@ pub async fn aggregate(
     let user_agent: &str = random_user_agent();
 
     // Add a random delay before making the request.
-    if random_delay || !debug {
+    if config.aggregator.random_delay || !config.debug {
         let nanos = SystemTime::now().duration_since(UNIX_EPOCH)?.subsec_nanos() as f32;
         let delay = ((nanos / 1_0000_0000 as f32).floor() as u64) + 1;
         tokio::time::sleep(Duration::from_secs(delay)).await;
--- a/src/results/aggregator.rs
+++ b/src/results/aggregator.rs
@@ -71,6 +71,7 @@ pub async fn aggregate(
     upstream_search_engines: &[EngineHandler],
     request_timeout: u8,
     safe_search: u8,
+    adaptive_window: bool,
 ) -> Result<SearchResults, Box<dyn std::error::Error>> {
     let client = CLIENT.get_or_init(|| {
         ClientBuilder::new()
@@ -78,6 +79,7 @@ pub async fn aggregate(
             .https_only(true)
             .gzip(true)
             .brotli(true)
+            .http2_adaptive_window(adaptive_window)
             .build()
             .unwrap()
     });
--- a/websurfx/config.lua
+++ b/websurfx/config.lua
@@ -14,8 +14,8 @@ rate_limiter = {
 	number_of_requests = 20, -- The number of request that are allowed within a provided time limit.
 	time_limit = 3, -- The time limit in which the quantity of requests that should be accepted.
 }
--- Set whether the server will use an adaptive/dynamic TCP window size
-adaptive_window = true
+-- Set whether the server will use an adaptive/dynamic HTTPS window size
+https_adaptive_window_size = false
 
 -- ### Search ###
 -- Filter results based on different levels. The levels provided are:
--- a/websurfx/config.lua
+++ b/websurfx/config.lua
@@ -14,6 +14,8 @@ rate_limiter = {
 	number_of_requests = 20, -- The number of request that are allowed within a provided time limit.
 	time_limit = 3, -- The time limit in which the quantity of requests that should be accepted.
 }
+-- Set whether the server will use an adaptive/dynamic TCP window size
+adaptive_window = true
 
 -- ### Search ###
 -- Filter results based on different levels. The levels provided are:
--- a/websurfx/config.lua
+++ b/websurfx/config.lua
@@ -14,7 +14,7 @@ rate_limiter = {
 	number_of_requests = 20, -- The number of request that are allowed within a provided time limit.
 	time_limit = 3, -- The time limit in which the quantity of requests that should be accepted.
 }
--- Set whether the server will use an adaptive/dynamic HTTPS window size
+-- Set whether the server will use an adaptive/dynamic HTTPS window size, see https://httpwg.org/specs/rfc9113.html#fc-principles
 https_adaptive_window_size = false
 
 -- ### Search ###
--- a/src/config/parser.rs
+++ b/src/config/parser.rs
@@ -30,6 +30,8 @@ pub struct Config {
     pub logging: bool,
     /// It stores the option to whether enable or disable debug mode.
     pub debug: bool,
+    /// It toggles whether to use adaptive TCP windows
+    pub adaptive_window: bool,
     /// It stores all the engine names that were enabled by the user.
     pub upstream_search_engines: HashMap<String, bool>,
     /// It stores the time (secs) which controls the server request timeout.
@@ -68,6 +70,7 @@ impl Config {
 
         let debug: bool = globals.get::<_, bool>("debug")?;
         let logging: bool = globals.get::<_, bool>("logging")?;
+        let adaptive_window: bool = globals.get::<_, bool>("adaptive_window")?;
 
         if !logging_initialized {
             set_logging_level(debug, logging);
@@ -125,6 +128,7 @@ impl Config {
             },
             logging,
             debug,
+            adaptive_window,
             upstream_search_engines: globals
                 .get::<_, HashMap<String, bool>>("upstream_search_engines")?,
             request_timeout: globals.get::<_, u8>("request_timeout")?,
--- a/src/config/parser.rs
+++ b/src/config/parser.rs
@@ -30,7 +30,7 @@ pub struct Config {
     pub logging: bool,
     /// It stores the option to whether enable or disable debug mode.
     pub debug: bool,
-    /// It toggles whether to use adaptive TCP windows
+    /// It toggles whether to use adaptive HTTP windows
     pub adaptive_window: bool,
     /// It stores all the engine names that were enabled by the user.
     pub upstream_search_engines: HashMap<String, bool>,
--- a/src/server/routes/search.rs
+++ b/src/server/routes/search.rs
@@ -218,6 +218,7 @@ async fn results(
                             .collect::<Vec<EngineHandler>>(),
                         config.request_timeout,
                         safe_search_level,
+                        config.adaptive_window,
                     )
                     .await?
                 }
--- a/src/server/routes/search.rs
+++ b/src/server/routes/search.rs
@@ -209,16 +209,13 @@ async fn results(
                     aggregate(
                         query,
                         page,
-                        config.aggregator.random_delay,
-                        config.debug,
+                        config,
                         &search_settings
                             .engines
                             .iter()
                             .filter_map(|engine| EngineHandler::new(engine).ok())
                             .collect::<Vec<EngineHandler>>(),
-                        config.request_timeout,
                         safe_search_level,
-                        config.adaptive_window,
                     )
                     .await?
                 }


websurfx/config.lua
```diff
@@ -14,6 +14,8 @@ rate_limiter = {
 	number_of_requests = 20, -- The number of request that are allowed within a provided time limit.
 	time_limit = 3, -- The time limit in which the quantity of requests that should be accepted.
 }
+-- Set whether the server will use an adaptive/dynamic TCP window size
+adaptive_window = true
```
```suggestion
-- Set whether the server will use an adaptive/dynamic HTTPS window size
https_adaptive_window_size = false
```

I would suggest renaming the config option to `https_adaptive_window_size` as I think it will better represent what the config option actually does, and also it may make it easier for the user to understand as well. Moreover, this option actually effects the `https` window size rather than the `tcp` window size. So I would suggest updating the documentation for the same as well. Additionally, I would also propose having these option disabled by default because we don't want to degrade the performance for users with unreliable network connection by keeping this option enabled by default. But instead allow the user to decide whether to enable/disable this option based on their network connections :slightly_smiling_face: .

neon-mmd (changes requested):
Thanks :heart: for working on this PR. We really appreciate it :+1: Keep it up!! But before we merge this PR, we would like to suggest a few changes. :slightly_smiling_face:

websurfx/config.lua
greengeckoz: Are you sure about it being an https window size. I can't find any information about that when searching, the only thing that comes up is the TCP window size.

neon-mmd: > Are you sure about it being an https window size. I can't find any information about that when searching, the only thing that comes up is the TCP window size.

Yes, because, you can already see it from the settings option `ClientBuilder` struct like it changes the window size for `https/http2` rather than the `tcp` window size. :slightly_smiling_face:

src/results/aggregator.rs
scottc943, line 1: minor complaint. I know we _could_ tell clippy to ignore the too many arguments warning, but I thinking this would be a good idea to help enforce a good design. If we are passing in multiple args, why not just pass it as a struct containing the arguments? https://stackoverflow.com/questions/67727687/passing-structures-as-arguments-in-rust

We could propose for a struct for aggragateparams like this:
``` rust
pub struct AggregationParams<'a> {
    pub query: &'a str,
    pub page: u32,
    pub random_delay: bool,
    pub debug: bool,
    pub upstream_search_engines: &'a [EngineHandler],
    pub request_timeout: u8,
    pub safe_search: u8,
    pub adaptive_window: bool,
}
```
----
Please note that the above struct is just an example and may not work exactly for this scenario

neon-mmd: > minor complaint. I know we _could_ tell clippy to ignore the too many arguments warning, but I thinking this would be a good idea to help enforce a good design. If we are passing in multiple args, why not just pass it as a struct containing the arguments? https://stackoverflow.com/questions/67727687/passing-structures-as-arguments-in-rust
> 
> We could propose for a struct for aggragateparams like this:
> 
> ```rust
> pub struct AggregationParams<'a> {
>     pub query: &'a str,
>     pub page: u32,
>     pub random_delay: bool,
>     pub debug: bool,
>     pub upstream_search_engines: &'a [EngineHandler],
>     pub request_timeout: u8,
>     pub safe_search: u8,
>     pub adaptive_window: bool,
> }
> ```
> 
> Please note that the above struct is just an example and may not work exactly for this scenario

I think maybe instead of having a seperate struct which would be very redundant, require reintialiating, passing the new struct and it will consume more memory. I think my suggestion would be to just pass the `config` struct directly as a reference. What do you think? Would it be a better idea? :slightly_smiling_face:

scottc943: I did miss the config struct existing so I’ll blame that on me not digging further through the code base. Yes, I do think the config struct would be better to pass in so there’s no need to repeat ourselves

greengeckoz: I altered aggregate() to take a &Config parameter to trim down several options

neon-mmd (approved):
Sorry for the late reply.
Thanks :heart: again for working on this. Since everything looks good. We will merge this PR as soon as possible. :rocket: :slightly_smiling_face:

PR #530, opened 2024-02-20
💄 Modernize the appearance of about page and make it responsive.

### LINKED ISSUE #433: 💄 Redesign the theme design for the `about` page
### What would you like to share?

Redesign the `simple` theme design for the `about` page for the search engine. To improve the design of the `simple` theme design for the `about` page. You will need to edit the code for the about page present in the `simple.css` file located in the `public/static/themes` folder located under the codebase (`websurfx` directory).

None

### Additional information

_No response_

cybrejon:
## What does this PR do?

This PR modifies `simple.css` to include better padding, margin, alignment and introduce RWD (responsive web design) to the about page.

### Small Screen Devices

127 0 0 1_8080_about

### Large Screen Devices

127 0 0 1_8080_about (1)

The current about's page is painfully outdated and not mobile-friendly. This hopes to fix that.

## Related issues

Closes #433


public/static/themes/simple.css
```diff
@@ -1,8 +1,42 @@
...
+  font-weight: 500;
+  font-display: swap;
+  src: url("https://fonts.gstatic.com/s/rubik/v28/iJWKBXyIfDnIV7nBrXw.woff2") format('woff2');
+}
+
+@font-face {
+  font-family: Rubik;
+  font-style: normal;
+  font-weight: 600;
+  font-display: swap;
```
neon-mmd:
```suggestion
@font-face {
    font-family: 'Rubik';
    font-style: normal;
    font-weight: 200 600;
    font-stretch: 0% 200%;
    font-display: swap;
    src: url(https://fonts.gstatic.com/s/rubik/v28/iJWKBXyIfDnIV7nErXyi0A.woff2) format('woff2');
    optional: sans-serif;
}
```

I think I would suggest instead of having separate font-faces for the same font. You can just provide one single font face as the font family you are using seems to be a variable font so you can effectively write it in one single `@font-face` attribute. I hope you got my point :slightly_smiling_face: 

Also, for more information on this, I would recommend the following resources to go through:
- https://stackoverflow.com/questions/28279989/multiple-font-weights-one-font-face-query

> [!Note]
> In the above link see the answer/reply with the title **Update 2022: use variable fonts**. 

- https://fonts.google.com/variablefonts?vfquery=rubik#axis-definitions

neon-mmd (changes requested):
Sorry for the late review.
Thanks :heart: for working on this PR. We really appreciate it :+1: Keep it up!! but before we merge this PR. We would suggest the following changes. :slightly_smiling_face:

public/static/themes/simple.css
cybrejon: Looks good, except for the the `optional` property, which is not passing CodeFactor nor is my css linter recommending that I push it. So does the single quotation marks around the font-family name, which is apparently unnecessary.

neon-mmd: > Looks good, except for the the `optional` property, which is not passing CodeFactor nor is my css linter recommending that I push it. So does the single quotation marks around the font-family name, which is apparently unnecessary.

Yes, they are not really required because I tested it on my side. It seems to work without quotation marks as well. So I would suggest removing it and also try running the app on your side too. :slightly_smiling_face:

cybrejon: Tested on my end just now, everything works.

neon-mmd (changes requested):
Sorry for the delay in the reply.
One last small suggestion, I feel it would be better to have all the `media queries` at the end of the file so that anyone else working on this future does not get confused or lost when searching/editing media queries. Moreover, I think it would make the code consistent too. What do you think? :slightly_smiling_face:

cybrejon:
> Sorry for the delay in the reply. One last small suggestion, I feel it would be better to have all the `media queries` at the end of the file so that anyone else working on this future does not get confused or lost when searching/editing media queries. Moreover, I think it would make the code consistent too. What do you think? 🙂

I believe they are already at the bottom-most part of the file.

neon-mmd:
> > Sorry for the delay in the reply. One last small suggestion, I feel it would be better to have all the `media queries` at the end of the file so that anyone else working on this future does not get confused or lost when searching/editing media queries. Moreover, I think it would make the code consistent too. What do you think? 🙂
> 
> I believe they are already at the bottom-most part of the file.

Ok I see, I am sorry, I think I overlooked it. Thanks :heart: for pointing it out. :slightly_smiling_face:

neon-mmd (approved):
Thanks :heart: again for working on this. Since everything looks good. We will merge this PR as soon as possible. :slightly_smiling_face:

PR #549, opened 2024-03-20
✨ Ranking of aggregated search results based on relevancy of the search result to the user's search query

### LINKED ISSUE #393: ✨ Ranking of aggregated search results based on relevancy of the search result to the user's search query
### Description

Implement an algorithm to rank (in other words, sort) the aggregated search results according to the relevancy of the search result based on the user's search query by analyzing three properties of the search results: 
- The search result's tile
- description
- and URL.

Also, checking each of the properties for how closely they are related if they are too closely related then it should be the first result in all search results. If it is least related, then it should be the last result to be shown in all the search results.

The algorithm should be implemented in the `aggregator.rs` file located in the `src/results/` folder under the codebase (`websurfx` directory) where the algorithm to aggregate search results lie.

### Screenshots

_No response_

None

### Additional information

_No response_

spencerjibz:
## What does this PR do?

Implements an algorithm to rank (in other words, sort) the aggregated search results according to the relevancy of the search result based on the user's search query by analyzing three properties of the search results:

- The search result's tile
- description
- and URL.

Also, check each of the properties for how closely they are related if they are too closely related then it should be the first result in all search results. If it is the least related, then it should be the last result to be shown in all the search results.

**Chosen Ranking function**:  [TF-IDF (Term Frequency-Inverse Document Frequency)](https://en.wikipedia.org/wiki/Tf%E2%80%93idf)
This is a weighting scheme often used by search engines as a central tool in scoring and ranking a document's [relevance](https://en.wikipedia.org/wiki/Relevance_(information_retrieval)) given a user query, find implementation details below:
-  Our implementation measures the relevance of a user's query in a collection of corpus (title, description and URL), then finally sorts the results by the relevance score.
- **Optional**:   we can also opt-in to weigh the query by synonyms found in the search title, description and URL. This feature can be enabled using the flag ``` use-synonyms-search```  or ``` use-non-static-synonyms-search```.

>  [!CAUTION]
> ***This feature flag enables using an offline thesaurus library (moby) to find synonyms and add them to the relevance calculations), there is both a performance and [binary size cost](https://docs.rs/thesaurus/0.5.2/thesaurus/#backend-comparison) when this feature is used.***. 

For more information on this ranking function, check out this [article](https://medium.com/analytics-vidhya/tf-idf-term-frequency-technique-easiest-explanation-for-text-classification-in-nlp-with-code-8ca3912e58c3)

Currently, the Search engine doesn't sort results by importance or relevance. This PR aims to solve that.

It can be tested by installing and running Websurfx as mentioned in the docs and on the readme and by launching the browser and thoroughly testing.  On searching for a topic, you find that the most relevant results are returned first . 

To try out using synonyms, build or run Websurfx with the feature flag``` use-synonyms-search``` 

## Related issues

Closes #393


greengeckoz (approved):
LGTM, the search is much more accurate now too.

Cargo.toml
```diff
@@ -13,45 +13,94 @@ bench = false
...
+    "std",
+], optional = true }
+cfg-if = { version = "1.0.0", default-features = false, optional = true }
+keyword_extraction = { version = "1.3.0", default-features = false, features = [
+    "tf_idf",
+
+
+] }
+
+stop-words = "0.8.0"
```
```suggestion
stop-words = { version = "0.8.0", default-features = false, features = [
     "iso",
]}
```
I would suggest disabling the default features for the `stop-words` crate as this will not enable any non-essential features when building/compiling the app, which can help reduce the build time significantly. :slightly_smiling_face:

Cargo.toml
```diff
@@ -13,45 +13,94 @@ bench = false
...
+thesaurus = { version = "0.5.2", default-features = false, optional = true, features = [
+    "wordnet",
+    "static",
+] }
```
```suggestion
thesaurus = { version = "0.5.2", default-features = false, optional = true, features = [
    "moby",
    "static",
] }
```

I would suggest using the `moby` synonym list instead as it contains support for other languages like `French`, `German`, etc. which can be useful when adding multi-language support in the future. I hope you got my point :slightly_smiling_face: .

Cargo.toml
```diff
@@ -81,10 +130,11 @@ strip = "symbols"
...
-redis-cache = ["dep:redis","dep:base64"]
-compress-cache-results = ["dep:async-compression","dep:cfg-if"]
-encrypt-cache-results = ["dep:chacha20poly1305","dep:chacha20"]
-cec-cache-results = ["compress-cache-results","encrypt-cache-results"]
+redis-cache = ["dep:redis", "dep:base64"]
+compress-cache-results = ["dep:async-compression", "dep:cfg-if"]
+encrypt-cache-results = ["dep:chacha20poly1305", "dep:chacha20"]
+cec-cache-results = ["compress-cache-results", "encrypt-cache-results"]
 experimental-io-uring = ["actix-web/experimental-io-uring"]
+use-synonyms-search = ["dep:thesaurus"]
```
```suggestion
use-synonyms-search = ["dep:thesaurus\static"]
use-non-static-synonyms-search = ["dep:thesaurus"]
```

I would suggest having two separate features for static and non-static synonyms search as this will allow devices with low-memory to also take advantage of the feature but also using lesser memory usage. :slightly_smiling_face: .

src/models/aggregation_models.rs
```diff
@@ -4,12 +4,13 @@
...
 use smallvec::SmallVec;
-
+#[cfg(feature = "use-synonyms-search")]
+use thesaurus::synonyms;
 /// A named struct to store the raw scraped search results scraped search results from the
 /// upstream search engines before aggregating it.It derives the Clone trait which is needed
 /// to write idiomatic rust using `Iterators`.
 /// (href url in html in simple words).
-#[derive(Clone, Serialize, Deserialize)]
+#[derive(Clone, Serialize, Deserialize, Debug)]
```
neon-mmd:
```suggestion
#[derive(Clone, Serialize, Deserialize)]
```

I think you forgot to remove the debug trait from here. So I would suggest removing it. :slightly_smiling_face: .

src/models/aggregation_models.rs
```diff
@@ -182,3 +225,54 @@ impl SearchResults {
...
+        total_score += tf_idf.get_score(token);
+    }
+
+    let result = total_score / (search_tokens.len() as f32);
+
+    if result.is_nan() {
+        0.0
+    } else {
+        result
+    }
```
neon-mmd:
```suggestion
    f32::from(!result.is_nan()) * result
```

I would suggest replacing the branched code with the equivalent branchless code, as this will reduce the CPU branch predictors misses and significantly improve the speed of the code.

neon-mmd (changes requested):
Sorry for the late review.

Thanks :heart: for working on this PR. We really appreciate it :+1: Keep it up!!  But before we merge this PR. We would like to suggest a few changes. :slightly_smiling_face:.

neon-mmd (approved):
Thanks :heart: again for working on this. Now since everything looks good. We will merge the PR as soon as possible. :slightly_smiling_face:

alamin655 (approved):

PR #555, opened 2024-04-20
🎨 Change `http` header word/sentence casing

tatodorov:
## What does this PR do?

The PR fixes the casing of the HTTP request headers sent to the search engines.

Currently all HTTP request headers are in snake case and upper case.  
The adopted standard is kebab case and title case.  

Here you will find more details 👉 [Standard request fields](https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Standard_request_fields)

Although using title case is optional, the usage of snake case for phrases is a must.

You can temporarily add `println!("{:#?}, header_map);` after each initialization of the HeaderMap.
Then just execute `cargo run` and make a search.

### DIFF
--- a/src/engines/bing.rs
+++ b/src/engines/bing.rs
@@ -79,13 +79,13 @@ impl SearchEngine for Bing {
         }
 
         let header_map = HeaderMap::try_from(&HashMap::from([
-            ("USER_AGENT".to_string(), user_agent.to_string()),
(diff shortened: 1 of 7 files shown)


neon-mmd (approved):
Sorry for the late review.

Thanks :heart: for working on this PR. We really appreciate it :+1: Good job!! Keep it up!! Since everything looks good. We will merge this PR as soon as possible :slightly_smiling_face: .

PR #573, opened 2024-05-25
✨ Option to use a proxy for outgoing `upstream search engine` requests

m00nwtchr:
## What does this PR do?

Adds a new config option to enable proxying of outgoing (to search engines) requests.

It allows for greater privacy of single-user/selfhosted instances by proxying requests via e.g. Tor.

1. Choose a proxy, e.g. tor
2. Set the proxy config value e.g.: `proxy = "socks5h://172.0.0.1:9050"`
3. Run websurfx


neon-mmd (changes requested):
Sorry for the late reply.

Thanks ❤️ for working on this issue. Good job!! Keep it up!! 👍 but we would like to suggest a few changes:

- Instead of using an empty string and stringified value for proxy option for config. We would suggest using conditionally compiled feature option of rust to allow the user to either enable/disable the option during the build stage as this will allow the user to for example not include the code for proxy in the final binary which can reduce the final binary size. Also, using this feature can increase the security of the server by disabling the proxy option and removing the code completely during the build stage. I hope you got our point 🙂 

#### Sample Code for the above suggestion

***Cargo.toml***

```diff
[package]
name = "websurfx"
version = "1.17.20"
edition = "2021"
description = "An open-source alternative to Searx that provides clean, ad-free, and organic results with incredible speed while keeping privacy and security in mind."
repository = "https://github.com/neon-mmd/websurfx"
license = "AGPL-3.0"

[[bin]]
name = "websurfx"
test = true
bench = false
path = "src/bin/websurfx.rs"

[dependencies]
reqwest = { version = "0.12.5", default-features = false, features = [
    "rustls-tls",
    "brotli",
    "gzip",
    "http2"
] }
tokio = { version = "1.32.0", features = [
    "rt-multi-thread",
    "macros",
    "fs",
    "io-util",
], default-features = false }
serde = { version = "1.0.209", default-features = false, features = ["derive"] }
serde_json = { version = "1.0.122", default-features = false }
maud = { version = "0.26.0", default-features = false, features = [
    "actix-web",
] }
scraper = { version = "0.20.0", default-features = false }
actix-web = { version = "4.9.0", features = [
    "cookies",
    "macros",
    "compress-brotli",
], default-features = false }
actix-files = { version = "0.6.5", default-features = false }
actix-cors = { version = "0.7.0", default-features = false }
fake-useragent = { version = "0.1.3", default-features = false }
env_logger = { version = "0.11.1", default-features = false }
log = { version = "0.4.21", default-features = false }
mlua = { version = "0.9.9", features = [
    "luajit",
    "vendored",
], default-features = false }
redis = { version = "0.25.4", features = [
    "tokio-comp",
    "connection-manager",
], default-features = false, optional = true }
blake3 = { version = "1.5.4", default-features = false }
error-stack = { version = "0.4.0", default-features = false, features = [
    "std",
] }
async-trait = { version = "0.1.80", default-features = false }
regex = { version = "1.9.4", features = ["perf"], default-features = false }
smallvec = { version = "1.13.1", features = [
    "union",
    "serde",
], default-features = false }
futures = { version = "0.3.30", default-features = false, features = ["alloc"] }
dhat = { version = "0.3.2", optional = true, default-features = false }
mimalloc = { version = "0.1.43", default-features = false }
async-once-cell = { version = "0.5.3", default-features = false }
actix-governor = { version = "0.5.0", default-features = false }
mini-moka = { version = "0.10", optional = true, default-features = false, features = [
    "sync",
] }
async-compression = { version = "0.4.12", default-features = false, features = [
    "brotli",
    "tokio",
], optional = true }
chacha20poly1305 = { version = "0.10.1", default-features = false, features = [
    "alloc",
    "getrandom",
], optional = true }
chacha20 = { version = "0.9.1", default-features = false, optional = true }
base64 = { version = "0.21.5", default-features = false, features = [
    "std",
], optional = true }
cfg-if = { version = "1.0.0", default-features = false, optional = true }
keyword_extraction = { version = "1.4.3", default-features = false, features = [
    "tf_idf",
] }

stop-words = { version = "0.8.0", default-features = false, features = ["iso"] }
thesaurus = { version = "0.5.2", default-features = false, optional = true, features = [
    "moby",
]}
itertools = {version = "0.13.0", default-features = false}

[dev-dependencies]
rusty-hook = { version = "^0.11.2", default-features = false }
criterion = { version = "0.5.1", default-features = false }
tempfile = { version = "3.10.1", default-features = false }

[build-dependencies]
lightningcss = { version = "1.0.0-alpha.57", default-features = false, features = [
    "grid",
] }
# Disabled until bug fixing update
# minify-js = { version = "0.6.0", default-features = false }
# Temporary fork with fix
minify-js = { git = "https://github.com/RuairidhWilliamson/minify-js", branch = "master", version = "0.6.0", default-features = false} 



[profile.dev]
opt-level = 0
debug = true
split-debuginfo = 'unpacked'
debug-assertions = true
overflow-checks = true
lto = false
panic = 'unwind'
incremental = true
codegen-units = 256
rpath = false

[profile.release]
opt-level = 3
debug = false # This should only be commented when testing with dhat profiler
# debug = 1 # This should only be uncommented when testing with dhat profiler
split-debuginfo = '...'
debug-assertions = false
overflow-checks = false
lto = 'thin'
panic = 'abort'
incremental = false
codegen-units = 1
rpath = false
strip = "symbols"

[profile.bsr1]
inherits = "release"
opt-level = "s"

[profile.bsr2]
inherits = "bsr1"
opt-level = "z"

[profile.lpcb1]
inherits = "release"
codegen-units = 16

[profile.lpcb2]
inherits = "lpcb1"
lto = "off"

[profile.lpcb3]
inherits = "lpcb2"
opt-level = 2

[profile.bsr_and_lpcb1]
inherits = "lpcb1"
opt-level = "s"

[profile.bsr_and_lpcb2]
inherits = "lpcb2"
opt-level = "s"

[profile.bsr_and_lpcb3]
inherits = "lpcb3"
opt-level = "s"

[profile.bsr_and_lpcb4]
inherits = "lpcb1"
opt-level = "z"

[profile.bsr_and_lpcb5]
inherits = "lpcb1"
opt-level = "z"

[profile.bsr_and_lpcb6]
inherits = "lpcb1"
opt-level = "z"

[features]
use-synonyms-search = ["thesaurus/static"]
default = ["memory-cache"]
+torified-search-queries = [reqwest/proxy]
dhat-heap = ["dep:dhat"]
memory-cache = ["dep:mini-moka"]
redis-cache = ["dep:redis", "dep:base64"]
compress-cache-results = ["dep:async-compression", "dep:cfg-if"]
encrypt-cache-results = ["dep:chacha20poly1305", "dep:chacha20"]
cec-cache-results = ["compress-cache-results", "encrypt-cache-results"]
experimental-io-uring = ["actix-web/experimental-io-uring"]
use-non-static-synonyms-search = ["thesaurus"]
```

Cargo.toml
```diff
@@ -174,7 +174,7 @@ opt-level = "z"
 
 [features]
 use-synonyms-search = ["thesaurus/static"]
-default = ["memory-cache"]
+default = ["memory-cache", "socks"]
```
```suggestion
default = ["memory-cache"]
```

I would suggest not having this as part of the default option because many users may not be running `tor` or may not want to use it for their own reasons. Also, by not having this in default we can leave it on the user to whether to use it or not like whether they want more privacy or not. I hope you got the idea. 🙂

src/config/parser.rs
```diff
@@ -6,6 +6,7 @@ use crate::handler::{file_path, FileType};
 use crate::models::parser_models::{AggregatorConfig, RateLimiter, Style};
 use log::LevelFilter;
 use mlua::Lua;
+use reqwest::Proxy;
```
neon-mmd, line 9:
```suggestion
#[cfg(feature = "socks")]
use reqwest::Proxy;
```

I would suggest having this behind a `cfg` feature macro as this will only add the following code to the binary if the `socks` feature has been enabled during build otherwise it will not include it thus reducing binary size and removing unneeded code. 🙂

src/config/parser.rs
```diff
@@ -48,6 +49,9 @@ pub struct Config {
     pub tcp_connection_keep_alive: u8,
     /// It stores the pool idle connection timeout in seconds.
     pub pool_idle_connection_timeout: u8,
+
+    /// Url of the proxy to use for outgoing requests.
+    pub proxy: Option<Proxy>,
```
neon-mmd:
```suggestion
    #[cfg(feature = "socks")]
    /// Url of the proxy to use for outgoing requests.
    pub proxy: Option<Proxy>,
```

I would suggest having this behind a `cfg` feature macro as this will only add the following code to the binary if the `socks` feature has been enabled during build otherwise it will not include it thus reducing binary size and removing unneeded code. 🙂

src/config/parser.rs
```diff
@@ -118,6 +122,15 @@ impl Config {
...
         };
 
+        let proxy_str = globals.get::<_, String>("proxy")?;
+        let proxy = match Proxy::all(proxy_str) {
+            Ok(proxy) => Some(proxy),
+            Err(_) => {
+                log::error!("Invalid proxy url, defaulting to no proxy.");
+                None
+            }
+        };
```
neon-mmd:
```suggestion
        #[cfg(feature = "socks")]
        let proxy_str = globals.get::<_, String>("proxy")?;
        #[cfg(feature = "socks")]
        let proxy = match Proxy::all(proxy_str) {
            Ok(proxy) => Some(proxy),
            Err(_) => {
                log::error!("Invalid proxy url, defaulting to no proxy.");
                None
            }
        };
```

I would suggest having this behind a `cfg` feature macro as this will only add the following code to the binary if the `socks` feature has been enabled during build otherwise it will not include it thus reducing binary size and removing unneeded code. 🙂

src/config/parser.rs
```diff
@@ -148,6 +161,7 @@ impl Config {
             safe_search,
             #[cfg(any(feature = "redis-cache", feature = "memory-cache"))]
             cache_expiry_time,
+            proxy,
```
neon-mmd, line 165:
```suggestion
            #[cfg(feature = "socks")]
            proxy,
```

I would suggest having this behind a `cfg` feature macro as this will only add the following code to the binary if the `socks` feature has been enabled during build otherwise it will not include it thus reducing binary size and removing unneeded code. 🙂

src/results/aggregator.rs
```diff
@@ -85,9 +85,13 @@ pub async fn aggregate(
...
-            .http2_adaptive_window(config.adaptive_window)
-            .build()
-            .unwrap()
+            .http2_adaptive_window(config.adaptive_window);
+
+        if config.proxy.is_some() {
+            cb = cb.proxy(config.proxy.clone().unwrap());
+        }
+
+        cb.build().unwrap()
```
neon-mmd, line 95:
```suggestion
            .http2_adaptive_window(config.adaptive_window);
        
        #[cfg(feature = "socks")]
        if config.proxy.is_some() {
            cb = cb.proxy(config.proxy.clone().unwrap());
        }

        cb.build().unwrap()
```

I would suggest having this behind a `cfg` feature macro as this will only add the following code to the binary if the `socks` feature has been enabled during build otherwise it will not include it thus reducing binary size and removing unneeded code. 🙂

websurfx/config.lua
```diff
@@ -73,3 +73,5 @@ upstream_search_engines = {
     Mojeek = false,
     Bing = false,
 } -- select the upstream search engines from which the results should be fetched.
+
+proxy = "" -- Proxy to send outgoing requests through. Set to empty string to disable.
```
```suggestion
proxy = "socks5h://127.0.0.1:9050" -- proxy string to send outgoing requests through. 
```

I would suggest changing this to use the localhost address with port 9050 by default and changing the explaination to exclude the empty string statement because the user can disable/enable this option using the `socks` feature during the build stage. 🙂

Cargo.toml
```diff
@@ -183,3 +183,4 @@ encrypt-cache-results = ["dep:chacha20poly1305", "dep:chacha20"]
 cec-cache-results = ["compress-cache-results", "encrypt-cache-results"]
 experimental-io-uring = ["actix-web/experimental-io-uring"]
 use-non-static-synonyms-search = ["thesaurus"]
+socks = ["reqwest/socks"]
```
neon-mmd, line 186:
```suggestion
torified-search-results = ["reqwest/socks"]
```

I think we should rename this feature to something like `torified-search-results` because the name `socks` sounds quite misleading for the user of the search engine like it could sound like the search engine supports running behind `tor` which is as you know not the case. So a name like `torified-search-results` will better explain the feature. What do you think? 🙂

neon-mmd (changes requested):
Thanks ❤️ for working on the previous suggestions and making some really cool changes but we would like to make a few more suggestions before we merge this PR in. 🙂

Cargo.toml
m00nwtchr: It's bad user experience to require people to compile their own version of the software just to enable one simple feature. The code in the project itself is fairly minimal, so the benefits of removing it would be pretty much 0. And you don't need to gate the whole proxy feature behind the flag, because `reqwest` supports http(s) proxies even without `reqwest/socks` enabled.

Cargo.toml
m00nwtchr: > because the name socks sounds quite misleading for the user of the search engine like it could sound like the search engine supports running behind tor which is as you know not the case. 

I have no idea what you're talking about. `socks` by itself has nothing to do with tor, and there are many other proxies/applications that make use of the protocol that have nothing to do with tor. And likewise there are other proxies that people may want to use *other* than tor.

Cargo.toml
neon-mmd: > It's bad user experience to require people to compile their own version of the software just to enable one simple feature. The code in the project itself is fairly minimal, so the benefits of removing it would be pretty much 0. And you don't need to gate the whole proxy feature behind the flag, because `reqwest` supports http(s) proxies even without `reqwest/socks` enabled.

Yes, true but I think if you think about security purposes and user customizability purposes having this feature compiled based is quite beneficial like if we don't have this compiled based then this code to parse remains in the app which means if a bad actor has access to the config file then they can manipulate it to use their malicious proxy address which is not great instead having the user to whether include this code or not can greatly reduce surface attack area like for example if the user does not compile the app with the proxy code then the bad actor even if they change it to their malicious address the app will never use it because it never has the code to. So this improves security by alot by reducing surface attack area. Also, security sometimes can be linked with privacy as well like if the server gets hacked by using this surface attack area this can also harm user privacy like the hacker could indeed have hold of sensitive data of the user not great again. Moreover, by having this as a compile only option improves user customizability too like they can include features they want otherwise they can exclude it during build time. So, that's why I would suggest having it as a compile/build only option. I hope you got the idea 🙂 .

neon-mmd: > It's bad user experience to require people to compile their own version of the software just to enable one simple feature. The code in the project itself is fairly minimal, so the benefits of removing it would be pretty much 0. And you don't need to gate the whole proxy feature behind the flag, because `reqwest` supports http(s) proxies even without `reqwest/socks` enabled.

After reconsidering your suggestion, I think you are right we should not block the user for such a small feature. So in accordance/agreement to your suggestions I will make a few new suggestions for the changes to make I hope you don't mind it 🙂 .

src/config/parser.rs
```diff
@@ -118,6 +122,15 @@ impl Config {
...
 
+        let proxy_str = globals.get::<_, String>("proxy")?;
+        let proxy = match Proxy::all(proxy_str) {
+            Ok(proxy) => Some(proxy),
+            Err(_) => {
+                log::error!("Invalid proxy url, defaulting to no proxy.");
+                None
+            }
+        };
+
```
neon-mmd:
```suggestion
                let proxy_opt = globals.get::<_, Option<String>>("proxy")?;
        let proxy = proxy_opt.and_then(|proxy_str| {
            Proxy::all(proxy_str).ok().and_then(|_| {
                log::error!("Invalid proxy url, defaulting to no proxy.");
                None
            })

```

I would suggest having the following parsing code which is more easy to read as well it uses `Option<String>` to parse the option from the config. So that we can have the following way for the user to disable or enable the option which I feel better conveys and documents like whether the option is disabled or not. While for any other unexpected cases like empty string or invalid proxy url can be handled via the above code with `Proxy::all`. 

**Disabled:**
```Lua
proxy = nil
```

**Enabled:**
```Lua
proxy = "socks5h://127.0.0.1:9050"
```

websurfx/config.lua
```suggestion
proxy = "" -- Proxy to send outgoing requests through. Set to empty string to disable.
```
```suggestion
proxy = nil -- Proxy to send outgoing requests through. Set to nil to disable.
```

As suggested previously, I would suggest having the value for this as nil which better explains/documents the user whether the option is disabled or enabled. 🙂

Cargo.toml
```suggestion
```

In accordance/agreement to your suggestion that we should not block this feature behind a compile flag then I would suggest removing this compile flag completely as we don't need to block such a small feature behind compile/build flag. 🙂

neon-mmd (changes requested):

neon-mmd:
@m00nwtchr it has been nearly one week, Any updates on this PR so far? Would love to hear :slightly_smiling_face:

m00nwtchr:
Alright, I merged in the changes, should be good to go now.

Cargo.toml
```diff
@@ -17,7 +17,7 @@ reqwest = { version = "0.12.5", default-features = false, features = [
     "rustls-tls",
     "brotli",
     "gzip",
-    "http2"
+    "http2",
```
neon-mmd, line 20:
```suggestion
    "http2",
    "socks",
```

Well, one last and minor suggestion we would like to make, I think you forgot to add the socks feature here. So I would suggest adding it. Other than everything looks good. :+1: :slightly_smiling_face: .

neon-mmd (changes requested):

neon-mmd (approved):
Thanks ❤️ again for working on such a cool PR. We really appreciate it. Good job!! Keep it up!! 👍. Since everything looks good. We will merge this PR as soon as possible 🙂 .

jfvillablanca (approved):

PR #599, opened 2024-09-02
🐛 Redis does not invalidate cached search results

### LINKED ISSUE #592: 🐛 Redis does not invalidate cached results
#### Work Expected From This Issue
The redis server does not invalidate the cached results even when the expiry time has been set for it. To fix this issue replace the `redis-rs's` `set_ex` command with `set_options` command and reimplement the fetching and caching of search results code in the `search.rs` file by caching only the uncached search results and not the cached results fetched from the redis server (this can be done using the flag based methodology via the results function in the same file).

The issue expects the following files to be changed/modified:
- `src/cache/redis_cacher.rs`
- `src/server/routes/search.rs`

> [!Note]
> All the files that are expected to be changed are located under the codebase (`websurfx` directory).

#### Reason Behind These Changes
The reason behind having these changes is to provide a fix for the redis cached search results invalidation issue and by providing this fix this can allow the cache to be invalidated on the right time and fetch the most current data without affecting user experience.

### Screenshots

_No response_

None

### Additional information

_No response_

neon-mmd:
## What does this PR do?

This PR fixes the issue regarding cached search results not invalidating in the redis server by reimplementing the caching code in `search` function and replacing the deprecated `set_ex` function with `set_options` function.

This change is essential because it fixes the issue regarding cache invalidation in redis server which if not dealt properly can cause cached results to never get invalidated and very old results to be provided to the users which can really provide a bad user experience.

It can be tested by installing and running Websurfx as mentioned in the `docs` and on the `readme` and by launching the browser and thoroughly testing. By checking one of the cached search results's ttl by requesting the same search query a couple of times. This can be done by running the following commands:

Connect a redis debugging client to the current server:
``` shell
redis-cli -p <port on which the server is running>
```
Fetch one of the cached search result:
``` shell
Key *
```
Checking the cached search result's ttl:
``` shell
tll <paste one of the keys copied from the above command as string>
```

## Author's checklist

- [x] Reimplement the caching code in the `search` function.
- [x] Replace the deprecated `set_ex` function with `set_options` function.
- [x] Bump the app version to `v1.17.20`. 

## Related issues

Closes #592

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -351,7 +351,7 @@ version = "0.4.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fec134f64e2bc57411226dfc4e52dec859ddfc7e711fc5e07b612584f000e4aa"
 dependencies = [
- "brotli 5.0.0",
+ "brotli",
  "flate2",
  "futures-core",
  "memchr",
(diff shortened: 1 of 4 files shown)


alamin655 (approved):

PR #621, opened 2024-10-12
♻️ Make the search bar work without `javascript` code

LaptopCat:
## What does this PR do?

This PR reduces usage of JavaScript on the frontend, instead replacing it with HTML features (using forms for searching) or server-side rendering (for pagination)

This change makes it possible to use basic features of websurfx (searching) without having JavaScript enabled on your browser. It also reduces the amount of scripts that the browser needs to load, making the frontend more lightweight.

Pull the changes, build websurfx and run it as usual. Then, open websurfx in your browser and disable JavaScript. Try searching for something, changing parameters (safesearch) and navigating result pages. It should work


neon-mmd (approved):
Sorry for the late reply.

We were actually planning to have these changes. So I would say this PR was made at the right time. Also, we really appreciate any PRs that reduce javascript usage by any means as you know javascript is always a nightmare for privacy. So a big Thanks ❤️ for working on such a wonderful PR. Good job!! Keep it up!! 👍 Since everything looks good. We will merge this PR as soon as possible. 🙂 .

spencerjibz (approved):

PR #633, opened 2024-11-27
✨ Provide `wikipedia` as a search engine option

gzsombor:
## What does this PR do?

Provide `wikipedia` as a search engine option.
Currently, it only search in the English Wikipedia, but it can be customized to use different ones. 
A better UI would be needed to enable configuring other languages 

Searching in Wikipedia is a very common thing

Run the app, select the Wikipedia engine, and search!

### DIFF
--- a/src/engines/bing.rs
+++ b/src/engines/bing.rs
@@ -15,6 +15,7 @@ use crate::models::engine_models::{EngineError, SearchEngine};
 
 use error_stack::{Report, Result, ResultExt};
 
+use super::common::build_cookie;
 use super::search_result_parser::SearchResultParser;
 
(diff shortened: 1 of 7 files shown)


neon-mmd (approved):
Sorry for the late reply.
Thanks ❤️ for working on such a wonderful PR. We really appreciate it. Keep it up!! The only thing we would suggest next time is to keep the changes a bit separated like we find there are some changes regarding refactor/restructuring though they are absolutely nice work but they look quite unrelated to what is being proposed that is addition of a new engine. Other than that since everything looks good. We will merge the PR as soon as possible 🙂 .

spencerjibz (approved):

PR #640, opened 2024-12-11
👽️ `Mojeek` parser update due to upstream breaking `api` changes

### LINKED ISSUE #631: :bug: Results are often prefixed with `localhost:port` which result in 404
### Description

On branch `rolling`.

Very often, when searching, the results page will show a number of links to websites/webpages which include both valid links and invalid ones--resulting in error 404.

The invalid links are prefixed with `http://localhost:port/` which of course results in 404.

One can just hover the mouse over the various results and see which links are valid and which are not.

This is definitely something which needs to be resolved.  Unless I my config file is misconfigured, creating the problem.

### Screenshots

_No response_

None

### Additional information

`config.lua` is [here](https://paste.xinu.at/NyKClQsE/)

jfvillablanca:
## What does this PR do?

- Fixes the CSS selector and parser for the search result `url` and `title`

- In order to access correct links from Mojeek search results

1. Check out to this branch
2. Enable `Mojeek` as an engine
3. Inspect the results 

#### Before:

image

#### After:

image

## Related issues

- Closes #631

### DIFF
--- a/src/engines/mojeek.rs
+++ b/src/engines/mojeek.rs
@@ -30,8 +30,8 @@ impl Mojeek {
             parser: SearchResultParser::new(
                 ".result-col",
                 ".results-standard li",
-                "a span.url",
-                "h2 a.title",
+                "h2 > a.title",
+                "a.ob",
                 "p.s",
             )?,
         })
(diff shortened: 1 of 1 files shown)


neon-mmd (approved):
Sorry for the late reply. There was a problem on my end during testing which caused this to happened.

Thanks ❤️ for the awesome PR. We really appreciate it. Since everything looks good. We will merge this PR right away 🚀 🙁 .

PR #675, opened 2025-04-23
🚸 Normalized urls for `yahoo` search engine to remove redirects

ROHANREDDYGADE:
What does this PR do?
This PR updates the way Yahoo URLs are handled. Instead of getting redirected Yahoo URLs, we now extract and use the normal, direct URLs.

Why is this change important?
Previously, URLs from Yahoo search results were in a redirected form. This change ensures that users receive the direct URLs instead of being redirected, improving accuracy and user experience.

How to test this PR locally?
   1. Clone the repository and check out this branch.

   2. Run the program and perform a search that would previously return a redirected Yahoo URL.

   3. Verify that the URLs returned are now the normal, direct URLs.

Previously how it works :-
Yahoo redirected urls

Now :-
Clean results

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,12 +44,12 @@ mlua = { version = "0.10.3", features = [
     "luajit",
     "vendored",
 ], default-features = false }
-redis = { version = "0.28.2", features = [
+redis = { version = "0.30.0", features = [
     "tokio-comp",
     "connection-manager",
     "tcp_nodelay"
 ], default-features = false, optional = true }
-blake3 = { version = "1.5.4", default-features = false }
+blake3 = { version = "1.8.2", default-features = false }
 error-stack = { version = "0.5.0", default-features = false, features = [
     "std",
 ] }
@@ -62,7 +62,7 @@ actix-governor = { version = "0.8.0", default-features = false }
 moka = { version = "0.12.8", optional = true, default-features = false, features = [
     "future",
 ] }
-async-compression = { version = "0.4.22", default-features = false, features = [
+async-compression = { version = "0.4.23", default-features = false, features = [
     "brotli",
     "tokio",
 ], optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,6 @@ reqwest = { version = "0.12.5", default-features = false, features = [
     "http2",
     "socks",
 ] }
-urlencoding = "2.1.3"
 tokio = { version = "1.43.1", features = [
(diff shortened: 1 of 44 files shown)


Cargo.toml
```diff
@@ -14,6 +14,7 @@ reqwest = { version = "0.12.5", default-features = false, features = [
     "http2",
     "socks",
 ] }
+urlencoding = "2.1.3"
```
neon-mmd: Is it possible to do this without using an extra dependency? I mean we are not against adding a dependency or the project goals have a fixed dependency limit but we would still prefer if this could be done without an extra dependency because this improves the compile/build time at the user's end. 🙂

neon-mmd (dismissed):
Sorry for the delay.

Thanks ❤️ for working on such a wonderful PR. We really appreciate it. Good job!! Keep it up!! 👍 . Since everything looks good. We will merge this PR as soon as possible 🙂 . 

> [!Note]
> I have left a short comment/concern on a code change. So if you could resolve then it would be great. Otherwise the PR is great. I really appreciate the cool work. 👍

Cargo.toml
ROHANREDDYGADE: Updated the Code: Removed urlencoding Dependency

I’ve made updates to the code to eliminate the need for the urlencoding crate. The functionality that previously relied on URL decoding now directly handles parsing and cleaning the URL without relying on the urlencoding library.

This change simplifies the dependencies and ensures the parsing of Yahoo redirect URLs is done natively without external crates for URL encoding/decoding.

Please let me know if you have any questions or further feedback!
yahoo

ROHANREDDYGADE:
✅ Replaced url::percent_encoding with a manual percent decoder to fix build errors.
All Yahoo engine URLs are now decoded manually using only the Rust standard library.

ROHANREDDYGADE:
The previous merge check was stuck on “Checking for the ability to merge automatically…” for over a day. To resolve this, I pushed a dummy commit to retrigger the CI workflows. I’ve verified that both build and test steps now run and pass successfully. ✅

Let me know if any further changes are needed!

jfvillablanca (dismissed):

neon-mmd:
@ROHANREDDYGADE 
Sorry for the ping.
It has been nearly week, any updates on this PR? 😅

ROHANREDDYGADE:
resolved those clippy format issues

spencerjibz (approved):

spencerjibz:
LGTM

PR #718, opened 2025-12-18
🐛 fix(handler): get `$HOME` environment variable at runtime

SchweGELBin:
## What does this PR do?

The `env!` macro gets an environment variable at **compile time**, emitting a compile time error it's not set.
The `std::env::var` function gets an environment variable at **runtime** and returns a result, which can be handled in the `unwrap_or` function.

This pull request replaces the first with the second approach.

It makes more sense to use the second variant as one can build the package on another system or environment that has a different HOME variable set.
This is espacially the case with prebuilt binaries or distribution provided packages.
On nix, the home variable would always evaluate to "/homeless-shelter", which isn't right and creates issues: One cannot use configs that are present in `$HOME/.config/websurfx/config.lua`

You can insert something like this under the changed line: `println!("Home: {}", home);`
Running `HOME=/newhome ./target/release/websurfx` after the build would print "HOME: /newhome" in the console instead of the home variable that was set during compilation.

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -51,7 +51,7 @@ static FILE_PATHS_FOR_DIFF_FILE_TYPES: OnceCell<Vec<(FileType, Vec<String>)>> =
 /// 2. Under project folder ( or codebase in other words) if it is not present
 ///    here then it returns an error as mentioned above.
 pub async fn file_path(file_type: FileType) -> Result<String, Error> {
-    let home = env!("HOME");
+    let home = std::env::var("HOME").unwrap_or(String::new());
 
     let file_path: Vec<String> = FILE_PATHS_FOR_DIFF_FILE_TYPES
         .get_or_init(|| async move {
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -51,7 +51,7 @@ static FILE_PATHS_FOR_DIFF_FILE_TYPES: OnceCell<Vec<(FileType, Vec<String>)>> =
 /// 2. Under project folder ( or codebase in other words) if it is not present
 ///    here then it returns an error as mentioned above.
 pub async fn file_path(file_type: FileType) -> Result<String, Error> {
-    let home = std::env::var("HOME").unwrap_or(String::new());
+    let home = std::env::var("HOME").unwrap_or_default();
 
     let file_path: Vec<String> = FILE_PATHS_FOR_DIFF_FILE_TYPES
         .get_or_init(|| async move {


spencerjibz:
@SchweGELBin, thanks for this contribution. I agree with you on having that flag check at runtime rather than compile time.
I had a similar issue when I initially contributed, which prompted me to run the project in WSL to check the builds. 
I wonder what @alamin655 or the others think?

SchweGELBin:
Is there a reason, why this isn't getting merged?
It is a very simple change that fixes a problem for some.

Have a great day!

spencerjibz:
> Is there a reason why this isn't getting merged? It is a very simple change that fixes a problem for some.
> 
> Have a great day!

It's a simple change indeed, since there is no response from other contributors, I'll merge it soon after fixing the failing tests.

neon-mmd:
@SchweGELBin Sorry for the ping and sorry for the delay in the reply.

I agree with you and @spencerjibz, I indeed see that having the check during runtime would be more viable then having it during the compile time. 

So from our side, it is a green signal and I would suggest on providing a fix for the same. 🙂 


Moreover, apart from the issue that was raised via this bug, I also see another issue that is if a user wants to change the `$HOME` variable then it will never would since the value from the variable will be embedded during compile time which will make it super inflexible and non-customizable which is really against the goal of the project as well. What do you think about this @spencerjibz @SchweGELBin ?? 😅

SchweGELBin:
Hey, I'm sorry, but I don't really understand your point from the previous comment.

Currently, the program reads the `$HOME` variable when the program gets compiled. Everytime, when someone runs the binary the variable will be the same.
With this patch, the program reads the `$HOME` variable when it first starts. The compilation doesn't matter here. You just need to restart the program if you want to change the variable.

Am I missing something?

In the mentioned PR in nixpkgs, the nix module starts a systemd service with the provided variable `HOME=/var/lib/websurfx`.
And with my testing it reads the config from `/var/lib/websurfx/.config/websurfx/config.lua` (I think, it's been a while), which isn't the prettiest path, but it at least works.

On nix, `$HOME` is set to "/homeless-shelter", so this change is pretty much necessary here if you want to use a custom config.

Have a great day!

neon-mmd:
> Currently, the program reads the $HOME variable when the program gets compiled. Everytime, when someone runs the binary the variable will be the same.

That's exactly what I was pointing out in my previous comment. 😅 

> Am I missing something?

No, you are already doing good. So nothing seems wrong. 

Also, a big thanks ❤️ on the aweaome work on supporting the nixos package. Like we were already wanting someone who could work in this area and keep maintaining it. Moreover this helps alot as it really improves the project and allows it to be supported on more platforms. So thanks again for the wonderful work!! 😍 .

Since everything looks good, we will approve and merge it as soon as possible. 🚀

src/handler.rs
```diff
@@ -51,7 +51,7 @@ static FILE_PATHS_FOR_DIFF_FILE_TYPES: OnceCell<Vec<(FileType, Vec<String>)>> =
 /// 2. Under project folder ( or codebase in other words) if it is not present
 ///    here then it returns an error as mentioned above.
 pub async fn file_path(file_type: FileType) -> Result<String, Error> {
-    let home = env!("HOME");
+    let home = std::env::var("HOME").unwrap_or(String::new());
```
neon-mmd: Everything looks good but before we merge it, we would suggest a small code style change, if you don't mind. 😉 

```suggestion
    let home = std::env::var("HOME").unwrap_or_default();
```

neon-mmd (changes requested):

spencerjibz:
LGMT, I have added your suggestion, let's merge this soon.

spencerjibz (approved):

neon-mmd:
> LGMT, I have added your suggestion, let's merge this soon.

Ok thanks ❤️ , Let's approve merge this now. 🚀 .

neon-mmd (approved):

SchweGELBin:
Thank you so much everyone! Have a great day!

PR #754, opened 2026-03-20
:sparkles: feat: add `json api` support for `search` endpoint

ghost:
## Summary

Adds an optional `format=json` query parameter to the `/search` endpoint. When provided, returns search results as structured JSON instead of rendering the HTML template.

## Usage

```bash
curl "http://localhost:8080/search?q=rust+programming&format=json"
```

## Response Format

```json
{
  "results": [
    {
      "title": "Rust Programming Language",
      "url": "https://rust-lang.org",
      "description": "...",
      "engine": ["duckduckgo"],
      "relevanceScore": 0.316
    }
  ],
  "engineErrorsInfo": [],
  "disallowed": false,
  "filtered": false,
  "safeSearchLevel": 2,
  "noEnginesSelected": false
}
```

## Changes

- Added `format` field to `SearchParams` struct
- Refactored `search` handler to extract shared logic into `fetch_results()` helper
- Single handler checks `format` param and returns either HTML or JSON
- Empty queries with `format=json` return HTTP 400 with `{"error": "Empty query provided"}`
- HTML responses are completely unchanged
- No new dependencies added

## Testing

Tested locally with `memory-cache` feature:
- ✅ JSON response returns valid, structured results
- ✅ HTML response unchanged
- ✅ Empty query returns 400 with error message
- ✅ Pagination (`page` param) works with both formats
- ✅ Safe search level respected in both formats
- ✅ Build succeeds with zero new warnings

## Summary by CodeRabbit

* **New Features**
  * Search endpoint accepts an optional json parameter to return JSON responses.

* **Bug Fixes**
  * When JSON is requested, malformed or empty queries now return a JSON 400; HTML behavior unchanged.
  * Improved error diagnostics for selector parsing to aid troubleshooting.

* **Refactor**
  * Response formatting and result fetching separated; caching and background population optimized to avoid unnecessary upstream work.

### DIFF
--- a/src/models/search_route.rs
+++ b/src/models/search_route.rs
@@ -6,7 +6,7 @@ use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 
 /// A named struct which deserializes all the user provided search parameters and stores them.
-#[derive(Deserialize)]
+#[derive(Deserialize, Debug)]
 pub struct SearchParams {
     /// It stores the search parameter option `q` (or query in simple words)
     /// of the search url.
--- a/src/models/search_route.rs
+++ b/src/models/search_route.rs
@@ -6,7 +6,7 @@ use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 
 /// A named struct which deserializes all the user provided search parameters and stores them.
-#[derive(Deserialize)]
+#[derive(Deserialize, Clone)]
 pub struct SearchParams {
     /// It stores the search parameter option `q` (or query in simple words)
     /// of the search url.
--- a/src/models/search_route.rs
+++ b/src/models/search_route.rs
@@ -6,7 +6,7 @@ use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 
 /// A named struct which deserializes all the user provided search parameters and stores them.
-#[derive(Deserialize, Clone)]
+#[derive(Deserialize)]
 pub struct SearchParams {
     /// It stores the search parameter option `q` (or query in simple words)
     /// of the search url.
--- a/src/models/search_route.rs
+++ b/src/models/search_route.rs
@@ -6,7 +6,7 @@ use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 
 /// A named struct which deserializes all the user provided search parameters and stores them.
-#[derive(Deserialize, Debug)]
+#[derive(Deserialize)]
 pub struct SearchParams {
     /// It stores the search parameter option `q` (or query in simple words)
     /// of the search url.
@@ -17,9 +17,9 @@ pub struct SearchParams {
     /// It stores the search parameter `safesearch` (or safe search level in simple words) of the
     /// search url.
     pub safesearch: Option<u8>,
-    /// It stores the search parameter `format` which determines the response format
-    /// (e.g., "json" for JSON API response).
-    pub format: Option<Cow<'static, str>>,
+    /// It stores the search parameter `json` which determines whether to return
+    /// a JSON API response instead of the HTML page.
+    pub json: Option<bool>,
 }
 
 /// A named struct which is used to deserialize the cookies fetched from the client side.
--- a/src/models/search_route.rs
+++ b/src/models/search_route.rs
@@ -17,6 +17,9 @@ pub struct SearchParams {
     /// It stores the search parameter `safesearch` (or safe search level in simple words) of the
     /// search url.
     pub safesearch: Option<u8>,
+    /// It stores the search parameter `format` which determines the response format
+    /// (e.g., "json" for JSON API response).
+    pub format: Option<Cow<'static, str>>,
 }
 
 /// A named struct which is used to deserialize the cookies fetched from the client side.
--- a/src/routes/search.rs
+++ b/src/routes/search.rs
@@ -45,16 +45,34 @@ static SHARED_CACHE: OnceCell<SharedCache> = OnceCell::const_new();
 /// # JSON API response
 /// curl "http://127.0.0.1:8080/search?q=sweden&format=json"
 /// ```
+/// Detect `format=json` from the raw query string before deserialization,
+/// so that parse failures can still return a JSON-shaped 400 response.
+fn is_json_request(query_string: &str) -> bool {
+    query_string
+        .split('&')
+        .any(|param| param.eq_ignore_ascii_case("format=json"))
+}
+
 #[get("/search")]
 pub async fn search(
     req: HttpRequest,
     config: web::Data<&'static Config>,
 ) -> Result<HttpResponse, Box<dyn std::error::Error>> {
-    let params = web::Query::<SearchParams>::from_query(req.query_string())?;
-    let json_mode = params
-        .format
-        .as_ref()
-        .map_or(false, |f| f.eq_ignore_ascii_case("json"));
+    // Detect JSON mode from raw query string first, so that parse failures
+    // still return a JSON-shaped error response instead of Actix's default 400.
+    let json_mode = is_json_request(req.query_string());
+
+    let params = match web::Query::<SearchParams>::from_query(req.query_string()) {
+        Ok(p) => p,
+        Err(e) => {
+            if json_mode {
+                return Ok(HttpResponse::BadRequest().json(serde_json::json!({
+                    "error": format!("Invalid query parameters: {}", e)
+                })));
+            }
+            return Err(e.into());
+        }
+    };
 
     let result = fetch_results(req, &config, params.into_inner()).await?;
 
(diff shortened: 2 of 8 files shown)


src/routes/search.rs
```diff
@@ -33,179 +34,226 @@ static SHARED_CACHE: OnceCell<SharedCache> = OnceCell::const_new();
...
+/// # JSON API response
+/// curl "http://127.0.0.1:8080/search?q=sweden&format=json"
 /// ```
+/// Detect `format=json` from the raw query string before deserialization,
+/// so that parse failures can still return a JSON-shaped 400 response.
+fn is_json_request(query_string: &str) -> bool {
+    query_string
+        .split('&')
+        .any(|param| param.eq_ignore_ascii_case("format=json"))
+}
```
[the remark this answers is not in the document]
neon-mmd: @metamorphosis-dev Any reasons for not working on this suggestion? If there is a valid reason for not working on it then we can close this as resolved. 😅

src/routes/search.rs
```diff
@@ -33,179 +34,226 @@ static SHARED_CACHE: OnceCell<SharedCache> = OnceCell::const_new();
...
+            // that is used for caching the results.
+            let tasks = pages
+                .iter()
+                .map(|page| results(config, &query_owned, *page, &search_settings, user_agent));
+            let fetched_results = futures::future::try_join_all(tasks).await?;
 
-        #[cfg(not(any(feature = "redis-cache", feature = "memory-cache")))]
-        {
-            current_results = results(&config, query, page, &search_settings, user_agent).await?;
+            tokio::spawn(async move { cache.cache_results(&fetched_results, &cache_keys).await });
```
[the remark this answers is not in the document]
neon-mmd: > _⚠️ Potential issue_ | _🟠 Major_
> 
> **Keep cache warming off the main request path.**
> 
> `current_results` is already available before this block, but the handler still awaits `try_join_all(tasks)` and bubbles any warm-up failure through `?`. A request can therefore stall or fail after the requested page was already resolved, and the `cached_results_exists` branch also refetches `pages` instead of only `cache_results_not_exists.1`, which adds unnecessary upstream load.
> 
> 🤖 Prompt for AI Agents

Resolved via the commit 9afd39c.

neon-mmd (changes requested):
Thanks ❤️ for working on such a wonderful PR. We really appreciate it. Good job!! Keep it up!! 👍 . 

But I would suggest a few suggestions before we merge this PR in. I hope you don't mind it. 😅 

- I would suggest replacing the match statements with only two branches to `if-let, else` style because this reduces [congnative code complexity](https://www.youtube.com/watch?v=CFRhGnuXG-4).

- I would suggest adding the `arguments`, `return`, etc section to the documentation of the `fetch_results` and `is_json_request` functions which will make it adhere to the standardized documentation style, the project uses across the whole codebase.

- I would suggest running the `cargo clippy --fix --all --all-features --allow-dirty` command at the end as it fixes any potential styling issues.

Other then that we would say, awesome PR 🚀  with alot of refactoring and with cool new feature implementation. Again, we really really appreciate the cool work you made.

ghost:
Thank you for the detailed review! Applied all suggestions — if-let over match for reduced branching, documentation added to both functions, and cargo clippy --fix run. Let me know if anything else needs adjusting.

src/routes/search.rs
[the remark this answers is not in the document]
neon-mmd: @coderabbitai could you explain the problem and suggestion in more depth?

src/models/search_route.rs
```diff
@@ -6,7 +6,7 @@ use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 
 /// A named struct which deserializes all the user provided search parameters and stores them.
-#[derive(Deserialize)]
+#[derive(Deserialize, Debug)]
```
neon-mmd:
```suggestion
#[derive(Deserialize)]
```

I would suggest removing this, as it is not required as explained in further suggestions ahead.

src/models/search_route.rs
```diff
@@ -17,6 +17,9 @@ pub struct SearchParams {
     /// It stores the search parameter `safesearch` (or safe search level in simple words) of the
     /// search url.
     pub safesearch: Option<u8>,
+    /// It stores the search parameter `format` which determines the response format
+    /// (e.g., "json" for JSON API response).
+    pub format: Option<Cow<'static, str>>,
```
neon-mmd:
```suggestion
    pub json: Option<bool>,
```

I would suggest on implementing an string-free implementation and more based on booleans as this will reduce the data type size as well as reduce the amount of edge cases to handle that is handling string sanitizations, checking for if the value is json or not, etc. And also, it will avoid complicated string handling like url encoding, etc as well when parsing it later on in the code which was also something explained in the previous suggestion by the automated coderabbitai review above.

src/routes/search.rs
```diff
@@ -17,6 +17,7 @@ use {crate::cache::SharedCache, tokio::sync::OnceCell};
 
 use actix_web::{HttpRequest, HttpResponse, get, http::header::ContentType, web};
 use regex::Regex;
+use serde_json;
```
neon-mmd:
```suggestion

```

No longer need for the unneeded import, so I would suggest removing it. 😅

src/routes/search.rs
```diff
@@ -33,179 +34,249 @@ static SHARED_CACHE: OnceCell<SharedCache> = OnceCell::const_new();
...
+/// # JSON API response
+/// curl "http://127.0.0.1:8080/search?q=sweden&format=json"
 /// ```
+/// Detect `format=json` from the raw query string before deserialization,
+/// so that parse failures can still return a JSON-shaped 400 response.
+fn is_json_request(query_string: &str) -> bool {
+    query_string
+        .split('&')
+        .any(|param| param.eq_ignore_ascii_case("format=json"))
+}
```
neon-mmd:
```suggestion

```

No longer the need for this function, as I would explain in the further suggestions. So I would suggest removing it as well.

src/routes/search.rs
```diff
@@ -33,179 +34,249 @@ static SHARED_CACHE: OnceCell<SharedCache> = OnceCell::const_new();
...
+
```
neon-mmd:
```suggestion
/// Handles the route of search page of the `websurfx` meta search engine website and it takes
/// two search url parameters `q` and `page` where `page` parameter is optional.
/// An optional `format` parameter can be provided to get results as JSON.
///
/// # Example
///
/// ```bash
/// # HTML response (default)
/// curl "http://127.0.0.1:8080/search?q=sweden&page=1"
///
/// # JSON API response
/// curl "http://127.0.0.1:8080/search?q=sweden&format=json"
/// ```
///
/// Detect `format=json` from the raw query string before deserialization,
/// so that parse failures can still return a JSON-shaped 400 response.

```

I would suggest providing the documentation for the search function.

src/routes/search.rs
```diff
@@ -33,179 +34,249 @@ static SHARED_CACHE: OnceCell<SharedCache> = OnceCell::const_new();
...
+        if json_mode {
+            return Ok(HttpResponse::BadRequest().json(serde_json::json!({
+                "error": "Empty query provided"
+            })));
+        }
+        Ok(HttpResponse::TemporaryRedirect()
+            .insert_header(("location", "/"))
+            .finish())
+    }
+}
```
neon-mmd:
```suggestion
    let params_result = web::Query::<SearchParams>::from_query(req.query_string());
    let params = if let Err(e) = params_result {
        if req
            .query_string()
            .split("&")
            .last()
            .unwrap()
            .contains("json")
        {
            return Ok(HttpResponse::BadRequest().json(serde_json::json!({
                "code": format!("{}",e.status_code()),
                "error": format!("Invalid query parameters: {}", e)
            })));
        }
        return Err(e.into());
    } else {
        params_result.unwrap()
    };

    let result = fetch_results(req, &config, &params).await?;

    if let Some((current_results, query, page)) = result {
        if let Some(json) = &params.json
            && *json
        {
            return Ok(HttpResponse::Ok()
                .content_type(ContentType::json())
                .json(&current_results));
        }
        Ok(HttpResponse::Ok().content_type(ContentType::html()).body(
            crate::templates::views::search::search(
                &config.style.colorscheme,
                &config.style.theme,
                &config.style.animation,
                &query,
                page,
                &current_results,
            )
            .0,
        ))
    } else {
        if let Some(json) = &params.json
            && *json
        {
            return Ok(HttpResponse::BadRequest().json(serde_json::json!({
                "code": "400",
                "error": "Empty query provided"
            })));
        }
        Ok(HttpResponse::TemporaryRedirect()
            .insert_header(("location", "/"))
            .finish())
    }
```

I would suggest reimplementing the logic something like this as this will avoid the complicated url decoding need and also avoids complicated edge case handling, etc. Also, it provides status codes on failures too. 

Additionally, the first `if-let-else` also avoids the need to derive debugs on any structs. I hope you get the idea. 😅

neon-mmd (changes requested):
I would suggest a few more changes then I think we can merge this in.

ghost:
Hi @neon-mmd! All remaining feedback has been addressed in the latest push:

- **search_route.rs**: Removed `#[derive(Debug)]` from `SearchParams`
- **search_route.rs**: Changed `format: Option<Cow<'static, str>>` → `json: Option<bool>` 
- **search_route.rs**: Updated doc comment accordingly
- **search.rs**: Removed `use serde_json;` import
- **search.rs**: Removed `is_json_request()` helper function entirely
- **search.rs**: Updated doc comment example to use `?json` instead of `?format=json`
- **search.rs**: Rewrote `search()` handler to follow your suggestion — raw query detection only on error path, `params.json` bool on success, `{"code": "400", "error": "..."}` JSON shape on errors

Pushed to `feature/json-api` branch. Ready for another look! 😅

ghost:
Update: addressed compilation errors too. Added `use actix_web::ResponseError`, added `Clone` derive to `SearchParams`, and fixed the `params` ownership issue. Build passes with `cargo build --all-features` (42 warnings, 0 errors). Ready for review!

ghost:
Fixes applied for the two latest CodeRabbit reviews:

**1. Cache hit performance (2985203587)**
Moved the random delay and `user_agent` lookup so they only execute on cache misses — warm cache hits now return immediately without any sleep.

**2. JSON detection robustness (2985215682)**
Replaced the fragile `split("&").last().contains("json")` with a proper segment-level check: `split('&').any(|seg| seg == "json" || seg.starts_with("json="))` — handles `?json&q=test` correctly and doesn't false-positive on `?q=json_programming`.

Build passes: 0 errors, 42 warnings.

neon-mmd (approved):
Since the PR conversations were having a lot of back and forth as well as alot of false positives from the coderabbitai. So we have marked the false positives review as resolved and I have worked on the changes required from my side. So with that the PR is ready to be merged with all the cool changes. And we will be merging this PR as soon as possible. 

Thanks again ❤️ for all the great work and working on this cool, awesome and superb PR. 🙂

PR #780, opened 2026-04-08
✨ feat(engine): provide the `sepiasearch` search engine

### LINKED ISSUE #778: ✨ `Sepiasearch` for the search engine
Provide the `qwant` search engine for the `websurfx` search engine.  To start working on this issue, you will need to create a new file with the name of the desired engine in the folder `src/engines/` which is located under the codebase (`websurfx` directory). Then in the file, just create a new struct with the name of the engine and then create two `impl` blocks for the struct. One for implementing the `SearchEngine` trait on it and the other for implementing the `parsers` to be used for aggregating results.

> [!NOTE]

ghost:
## What does this PR do?

Adds **SepiaSearch** (the PeerTube video search index) as a new upstream search engine. This is the first engine in websurfx that consumes a JSON API rather than scraping HTML.

**Changes:**
- New engine file `src/engines/sepiasearch.rs` querying `https://sepiasearch.org/api/v1/search/videos` and parsing the JSON response using serde
- Registered in `src/engines/mod.rs`, the engine match statement in `src/models/engine.rs`, and `websurfx/config.lua`
- Added `form_urlencoded` as a direct dependency (already present as a transitive dep via reqwest) for proper query encoding

**Implementation details:**
- Uses the existing `fetch_json_as_bytes_from_upstream` trait method (previously defined but unused by any engine)
- Pagination via `start`/`count` params (0-based offset, 10 results per page)
- Safe search mapped to the `nsfw` API param: level 0 → `"both"`, levels 1+ → `"false"` (matching the SearXNG reference implementation)
- API error responses are surfaced rather than silently masked as empty results
- Missing `description` fields gracefully default to empty string

SepiaSearch is the largest federated video search index, aggregating content from thousands of PeerTube instances. Adding it gives websurfx users access to video results from the fediverse, content that is not available through any of the currently supported engines (DuckDuckGo, Brave, SearX, etc.).

1. Enable SepiaSearch in `websurfx/config.lua`:
   ```lua
   SepiaSearch = true,
   ```
2. Build and run:
   ```shell
   cargo run
   ```
3. Search for any query with SepiaSearch enabled. Video results from PeerTube instances should appear.
4. Run unit tests:
   ```shell
   cargo test --lib engines::sepiasearch
   ```

## Author's checklist

- [x] Follows existing engine patterns (struct + `SearchEngine` trait impl + constructor)
- [x] All private struct fields have doc comments (clippy `missing_docs_in_private_items`)
- [x] No unnecessary `Content-Type` header on GET requests
- [x] Query parameter is properly URL-encoded
- [x] API errors are detected and surfaced with context
- [x] No new transitive dependencies introduced (`form_urlencoded` was already in the lockfile)
- [x] Unit tests for normal response, missing optional fields, and empty results

## Related issues
Closes #778 

## Summary by CodeRabbit

* **New Features**
  * Added SepiaSearch as a new upstream search engine option with a configurable toggle in settings.

* **Chores**
  * Added form_urlencoded dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ name = "websurfx"
 path = "src/main.rs"
 
 [dependencies]
+form_urlencoded = "1.2"
 rayon = "1.10.0"
 arc-swap = "1.9.0"
 reqwest = { version = "0.12.19", default-features = false, features = [
--- a/src/engines/sepiasearch.rs
+++ b/src/engines/sepiasearch.rs
@@ -79,11 +79,7 @@ impl SearchEngine for SepiaSearch {
         client: &Client,
         safe_search: u8,
     ) -> Result<Vec<(String, SearchResult)>, EngineError> {
-        // SepiaSearch uses nsfw parameter: 0 -> "both", 1 -> "false", 2 -> "false"
-        let nsfw = match safe_search {
-            0 => "both",
-            _ => "false",
-        };
+        let nsfw = if safe_search == 0 { "both" } else { "false" };
 
         // Pagination: 0-based offset, 10 results per page
         let start = page * 10;
@@ -98,6 +94,10 @@ impl SearchEngine for SepiaSearch {
             (
                 "Referer".to_string(),
                 "https://sepiasearch.org/".to_string(),
+            ),
+            (
+                "Content-Type".to_string(),
+                "text/html; charset=utf-8".to_string(),
             ),
         ]))
         .change_context(EngineError::UnexpectedError)?;
--- a/src/engines/sepiasearch.rs
+++ b/src/engines/sepiasearch.rs
@@ -1,0 +1,173 @@
+//! The `sepiasearch` module handles the fetching of results from the SepiaSearch video search
+//! engine (PeerTube search index) by querying its JSON API with the user provided query.
+
+use reqwest::{Client, header::HeaderMap};
+use serde::Deserialize;
+use std::collections::HashMap;
+
+use crate::models::aggregation::SearchResult;
+use crate::models::engine::{EngineError, SearchEngine};
+use error_stack::{Report, Result, ResultExt};
+
+/// A new SepiaSearch engine type defined in-order to implement the `SearchEngine` trait.
+pub struct SepiaSearch;
+
+/// The JSON response structure returned by the SepiaSearch API.
+#[derive(Deserialize)]
+struct SepiaSearchResponse {
+    /// The list of video results returned by the API.
+    data: Vec<SepiaSearchVideo>,
+    /// The total number of results available.
+    total: Option<u32>,
+    /// An error message, if the API returned one.
+    error: Option<String>,
+}
+
+/// A single video result from the SepiaSearch API.
+#[derive(Deserialize)]
+struct SepiaSearchVideo {
+    /// The title of the video.
+    name: String,
+    /// The URL to watch the video.
+    url: String,
+    /// An optional description of the video.
+    description: Option<String>,
+}
+
+impl SepiaSearch {
+    /// Creates a new SepiaSearch engine instance.
+    pub fn new() -> Result<SepiaSearch, EngineError> {
+        Ok(Self)
+    }
+
+    /// Parses the raw JSON response body into a list of search results.
+    fn parse_json_response(json: &[u8]) -> Result<Vec<(String, SearchResult)>, EngineError> {
+        let response: SepiaSearchResponse =
+            serde_json::from_slice(json).change_context(EngineError::UnexpectedError)?;
+
+        if let Some(err) = &response.error {
+            return Err(Report::new(EngineError::UnexpectedError)
+                .attach(format!("SepiaSearch API error: {err}")));
+        }
+
+        let results = response
+            .data
+            .into_iter()
+            .map(|video| {
+                let description = video.description.unwrap_or_default().trim().to_string();
+                let search_result = SearchResult::new(
+                    video.name.trim(),
+                    video.url.as_str(),
+                    description.as_str(),
+                    &["sepiasearch"],
+                );
+                (search_result.url.clone(), search_result)
+            })
+            .collect();
+
+        Ok(results)
+    }
+}
+
+#[async_trait::async_trait]
+impl SearchEngine for SepiaSearch {
+    async fn results(
+        &self,
+        query: &str,
+        page: u32,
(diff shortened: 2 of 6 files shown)


src/engines/sepiasearch.rs
```diff
@@ -0,0 +1,173 @@
...
+        safe_search: u8,
+    ) -> Result<Vec<(String, SearchResult)>, EngineError> {
+        // SepiaSearch uses nsfw parameter: 0 -> "both", 1 -> "false", 2 -> "false"
+        let nsfw = match safe_search {
+            0 => "both",
+            _ => "false",
+        };
+
+        // Pagination: 0-based offset, 10 results per page
+        let start = page * 10;
```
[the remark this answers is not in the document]
neon-mmd: False negative. So sorry closing it as such. Thanks :heart: for the review.

src/engines/sepiasearch.rs
```diff
@@ -0,0 +1,173 @@
...
+        );
+
+        let header_map = HeaderMap::try_from(&HashMap::from([
+            ("User-Agent".to_string(), user_agent.to_string()),
+            (
+                "Referer".to_string(),
+                "https://sepiasearch.org/".to_string(),
+            ),
+        ]))
+        .change_context(EngineError::UnexpectedError)?;
```
neon-mmd, line 103:
```suggestion
        let header_map = HeaderMap::try_from(&HashMap::from([
            ("User-Agent".to_string(), user_agent.to_string()),
            (
                "Referer".to_string(),
                "https://sepiasearch.org/".to_string(),
            ),
            ("Content-Type".to_string(), "text/html; charset=utf-8".to_string()
            ),
        ]))
        .change_context(EngineError::UnexpectedError)?;
```

I would suggest adding the `content-type` header because this ensures that the user doesn't get marked as a bot when requesting the data from upstream.

src/engines/sepiasearch.rs
```diff
@@ -0,0 +1,173 @@
...
+        page: u32,
+        user_agent: &str,
+        client: &Client,
+        safe_search: u8,
+    ) -> Result<Vec<(String, SearchResult)>, EngineError> {
+        // SepiaSearch uses nsfw parameter: 0 -> "both", 1 -> "false", 2 -> "false"
+        let nsfw = match safe_search {
+            0 => "both",
+            _ => "false",
+        };
```
neon-mmd:
```suggestion
        let nsfw = if safe_search == 0 { "both" } else { "false" };
        
```
I would suggest using `if-else` then using `match` block as it reduces cognitive code complexity. You know what I mean. 😅

src/engines/sepiasearch.rs
```diff
@@ -0,0 +1,173 @@
...
+            serde_json::from_slice(json).change_context(EngineError::UnexpectedError)?;
+
+        if let Some(err) = &response.error {
+            return Err(Report::new(EngineError::UnexpectedError)
+                .attach(format!("SepiaSearch API error: {err}")));
+        }
+
+        let results = response
+            .data
+            .into_iter()
```
neon-mmd, line 55:
```suggestion
            .into_par_iter()
```
Maybe you can improve the speed via rayon?

Cargo.toml
```diff
@@ -11,6 +11,7 @@ name = "websurfx"
 path = "src/main.rs"
 
 [dependencies]
+form_urlencoded = "1.2"
```
neon-mmd, line 14: Any strong reasons to have this? I mean like if we didn't have it then the whole search thing will fail?

neon-mmd (changes requested):
Thanks ❤️ for working on this PR. We really appreciate it. But before we merge this in, we would request you to do some changes, if you don't mind.

---
Moreover, a little offtopic but if you do remain consistent in contributing upstream engines and maintaining them, then we will be adding you as a candidate as an engine maintainer for the project. Also, feel free to join our community discord server and have a chat with us.

ghost:
Thanks for the review and the kind words! I've pushed a commit with the Content-Type header and the if-else change.

Regarding rayon, I kept `into_iter()` for now. SepiaSearch only returns 10 results per page, so the parallelization overhead would probably outweigh any gains. The other engines also use sequential iteration for result mapping, and rayon is used later in the aggregator where it processes combined results across all engines. But happy to switch if you'd prefer!

On `form_urlencoded`, fair question. It's not strictly required since the other engines just use `format!()` without encoding. I included it because it's already a transitive dependency through reqwest (so zero compile cost), and it handles edge cases like queries containing `&` or `#`. But if you'd rather I match the existing pattern and drop it, totally fine, just say the word.

src/engines/sepiasearch.rs
neon-mmd: Fixed in f2a6f19d9f9862bbc33fc829a2e2f172d5553426.

src/engines/sepiasearch.rs
neon-mmd: Fixed in f2a6f19d9f9862bbc33fc829a2e2f172d5553426.

neon-mmd:
> Thanks for the review and the kind words! I've pushed a commit with the Content-Type header and the if-else change.

Good work :white_check_mark: .

> Regarding rayon, I kept `into_iter()` for now. SepiaSearch only returns 10 results per page, so the parallelization overhead would probably outweigh any gains. The other engines also use sequential iteration for result mapping, and rayon is used later in the aggregator where it processes combined results across all engines. But happy to switch if you'd prefer!

Makes sense, if it is just 10 results, then we surely don't need it. It will just be an overkill, nothing much.

> On `form_urlencoded`, fair question. It's not strictly required since the other engines just use `format!()` without encoding. I included it because it's already a transitive dependency through reqwest (so zero compile cost), and it handles edge cases like queries containing `&` or `#`. But if you'd rather I match the existing pattern and drop it, totally fine, just say the word.

Ok, I think maybe, if this could be done via the UI then, that would be better I feel, like it will be universal for all engines (but this would be a little bit of an exploration work on how it can be achieved like without using Javascript.). Or if this could be done at more top-level like in the search route that would be nice as well because as you mentioned like it can handle edge cases for special characters. So this can be worked on in a separate PR and issue. I will open an issue for it once I am on my machine for sure. 

So with that we have marked all the reviews as resolved.

neon-mmd (approved):
Thanks ❤️ again for all the awesome work. Since everything looks good now. Let's merge this in. :star_struck:

